PR #10, opened 2018-03-31
Fix sorting versions for IEDriver

hron:


hron:
FYI, the fails on CI are unrelated to the requested change.

titusfortner:
Thanks for submitting this. Upon further review, I see that I'm not even matching the right part of the String for IE in the first place. I also don't know why I'm converting things to `.to_f`
Unfortunately, testing out this code, the result still isn't sorted correctly.

I think I want to change the float compares to `Gem::Version` compares since it is already available code that properly handles all of the non-padded/minor/point issues.

hron:
@titusfortner please check the improved version.

lib/webdrivers/iedriver.rb
```diff
@@ -35,7 +41,7 @@ def downloads
           items = doc.css("Key").collect(&:text)
           items.select! { |item| item.include?('IEDriverServer_Win32') }
           ds = items.each_with_object({}) do |item, hash|
-            key = normalize item[/^[^\/]+/]
+            key = normalize item[/\d+\.\d+\.\d+(?=\.zip$)/]
```
titusfortner: I'm thinking we want to avoid the digit formatting and grab everything before the `_` like: `item[/([^_]+)\.zip/, 1]`

titusfortner:
Let me know about changing that Regex, otherwise, this looks good.

kapoorlakshya:
@hron - Eagerly waiting for you to update the PR and have this fix merged :)

hron:
@titusfortner @kapoorlakshya okay! Sorry for the delay ;)

kapoorlakshya:
Thanks, @hron! 👍

PR #11, opened 2018-04-14
MicrosoftWebDriver - Implement version checking and download of latest binary.

kapoorlakshya:
Hi @titusfortner,

I have updated the `MSWebdriver` implementation.  Earlier it would raise an error asking the user to manually download the binary. With this PR it now queries the current Edge version and downloads the corresponding binary from the Microsoft developer website if an existing binary does not exist. 

## Caveat

1. Unfortunately, because `MicrosoftWebDriver.exe` does not provide a `--version` option, there is no way to verify if an existing binary is the correct version or not. 

```
USAGE:
        MicrosoftWebDriver [OPTIONS]

OPTIONS:
        --host=<HostName>    Host IP to use for the WebDriver server (default: localhost)
        --port=<PortNumber>  Port to use for the WebDriver server (default: 17556)
        --package=<Package>  ApplicationUserModelId (AUMID) for the application to be launched by the WebDriver server
        --verbose            Outputs requests received and responses sent by the WebDriver server
        --silent             Outputs nothing
```

The implementation simply assumes the existing binary is the expected version and returns the `location`. If the existing binary is incompatible, the user will see the following from Selenium:

```
unable to connect to MicrosoftWebDriver localhost:17556
```

If we proceed with this, we'll have to advise the user to delete the existing binary and retry.

2. The actual download page for the binary has a redirect on it. This kicks in the `Net::HTTPRedirection` case in `get()` on `lib/webdrivers/common.rb:80` and uses the `location` attribute. The `location` returned in the response is an invalid URL pointing to `/en-us/`. To workaround this, I have left the `base_url` as is and added `download_page` to return the actual download page URL.

Let me know what you think of this implementation. I will push the updated specs soon and squash all commits before the final approval.


lib/webdrivers/common.rb
```diff
@@ -9,18 +9,24 @@ def update
...
+        return location if released.nil? && binary_exists
 
         if released.nil?
           msg = "Unable to find the latest version of #{file_name}; try downloading manually from #{base_url} and place in #{install_dir}"
           raise StandardError, msg
         end
 
-        return location if current == released
-        remove && download
+        if current == released && binary_exists # Already have latest/matching one
```
titusfortner: When would `binary_exist` be false? If it doesn't exist, wouldn't `current` throw an exception?

lib/webdrivers/mswebdriver.rb
```diff
@@ -23,32 +28,30 @@ def file_name
...
-          Webdrivers.logger.debug "Attempting to Download Latest Insider's Version"
-          "https://download.microsoft.com/download/1/4/1/14156DA0-D40F-460A-B14D-1B264CA081A5/MicrosoftWebDriver.exe"
-        end
+        available = {} # version => download_url
+        doc       = Nokogiri::HTML(get(download_page))
+        doc.xpath("//li[@class='driver-download']/a").map { |link|
+          next if link.text == 'Insiders' # Ignore Insiders release
+          available[link.text.scan(/\d+/).first.to_i] = link['href']
+        }
+        available
```
titusfortner: This code is good, but an interesting, more compact alternative to consider in the future:
```
array = Nokogiri::HTML(get(download_page)).xpath("//li[@class='driver-download']/a")
array.each_with_object({}) do |link, hash|
  next if link.text == 'Insiders'
  hash[link.text.scan(/\d+/).first.to_i] = link['href']
end
```

lib/webdrivers/mswebdriver.rb
```diff
@@ -23,32 +28,30 @@ def file_name
...
       def base_url
         'https://www.microsoft.com/en-us/download'
       end
 
+       # Using this as base_url yields Net::HTTPRedirection with a partial location, which fails
+       # on the second Common#get call in lib/webdrivers/common.rb:77.
+       # Using existing base_url to get past that hurdle until we discuss a better way to handle
+       # the HTTP status.
+      def download_page
+        "https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/"
```
titusfortner: Can this page be the base_url? Not sure what the distinction is, even with your comment.

titusfortner (commented):
Just a couple questions, but this looks good.

lib/webdrivers/common.rb
kapoorlakshya: Unlike other classes `current` is getting the browser version (Edge) from the Windows registry instead of the binary version. This means that `current` will still return a version when `binary_exists` is false. This is to work around the limitation of `MicrosoftWebDriver.exe` not having a `version` argument to retrieve the webdriver version. 

The logic here is to retrieve the Edge version through `current` and then download the corresponding binary if a binary file (any version) does not exist.

lib/webdrivers/mswebdriver.rb
kapoorlakshya: I noticed that approach in other classes, but I couldn't really wrap my head around `each_with_object` so I took the simpler approach. I will update the PR to with the more elegant solution provided by you 👍.

lib/webdrivers/mswebdriver.rb
kapoorlakshya: Huh... so I was running into issues  with HTTP redirection when using the `download_page` URL. I wonder if it was because I was behind a proxy at work when I worked on this, but now that I am at home I am not experiencing it. I will have to test this out on Monday. For now, I have removed `download_page` and updated `base_url`.

kapoorlakshya:
Argh... I messed up a rebase. Will submit another PR with changes discussed here.

kapoorlakshya:
Argh... I pushed a messed up rebase and GitHub closed this PR. I am opening a new PR with discussed changes to avoid this mess.

lib/webdrivers/mswebdriver.rb
titusfortner: it just avoids the necessity of defining it, then using it in a loop, then returning it by putting it all in one step

PR #13, opened 2018-04-29
MSWebdrivers now uses Edge version to download the correct binary version.

kapoorlakshya:
Hi @titusfortner,

I have updated the `MSWebdriver` implementation.  Earlier it would raise an error asking the user to manually download the binary. With this PR it now queries the current Edge version and downloads the corresponding binary from the Microsoft developer website if an existing binary does not exist.

I have added a spec to verify edge version is retrieved and all existing specs still pass.

## Caveat

Unfortunately, because `MicrosoftWebDriver.exe` does not provide a `--version` option, there is no way to verify if an existing binary is the correct version or not. We instead use the Edge version from Windows Registry to download the corresponding binary, if one does not already exist in `.webdrivers`.

```
USAGE:
        MicrosoftWebDriver [OPTIONS]

OPTIONS:
        --host=<HostName>    Host IP to use for the WebDriver server (default: localhost)
        --port=<PortNumber>  Port to use for the WebDriver server (default: 17556)
        --package=<Package>  ApplicationUserModelId (AUMID) for the application to be launched by the WebDriver server
        --verbose            Outputs requests received and responses sent by the WebDriver server
        --silent             Outputs nothing
```

If a binary already exists, the implementation simply assumes it is the expected version and returns the `location`. If the existing binary is incompatible, the user will see the following from Selenium:

```
unable to connect to MicrosoftWebDriver localhost:17556
```

If we proceed with this, we'll have to advise the user to delete the existing binary and retry. Let me know what you think of this implementation.

### DIFF
--- a/spec/mswebdriver_spec.rb
+++ b/spec/mswebdriver_spec.rb
@@ -19,7 +19,7 @@ describe Webdrivers::MSWebdriver do
     before { allow(mswebdriver).to receive(:site_available?).and_return(false) }
 
     it 'raises exception downloading' do
-      expect {mswebdriver.download}.to raise_error(StandardError, "Can not reach site")
+      expect { mswebdriver.download }.to raise_error(StandardError, "Can not reach site")
     end
   end
 
--- a/lib/webdrivers/common.rb
+++ b/lib/webdrivers/common.rb
@@ -9,18 +9,24 @@ module Webdrivers
         unless site_available?
           return current.nil? ? nil : binary
         end
-        released = latest()
-        location = binary()
+        released      = latest()
+        location      = binary()
+        binary_exists = File.exists?(location)
 
-        return location if released.nil? && File.exist?(location)
+        return location if released.nil? && binary_exists
 
         if released.nil?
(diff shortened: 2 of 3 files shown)


titusfortner:
I'm not going to have time to test this on Windows, so let me know if there is an issue at your work computer tomorrow. If not, then I'll merge and release the update. Thanks.

spec/mswebdriver_spec.rb
```diff
@@ -4,6 +4,10 @@
 
   let(:mswebdriver) { Webdrivers::MSWebdriver }
 
+  it "retrieves edge version" do
+    expect(mswebdriver.current).to be >= 16299
+  end
+
```
titusfortner: `mswebdriver.current` will always fail on Travis, it being Linux and all :)
I think we can just delete this spec

kapoorlakshya: Sounds good!

kapoorlakshya:
The `base_url` seems work work fine through proxy. I will remove the spec and push a commit, after which you can merge the PR. Want me to bump the version in this PR or you want to do it yourself?

Issue #14, opened 2018-05-02
MSWebdriver ignores currently installed version of Edge and fetches the latest available binary.


kapoorlakshya:
I am already working on a PR to fix this, but submitting this here in case anyone runs into this issue.

Microsoft recently released a new webdriver version 6.17134 for Edge 17.17134. If you are using this gem to launch an older version, say 16.16299, Selenium will fail to launch the browser with this error:

```
#<Errno::ECONNREFUSED: Failed to open TCP connection to localhost:17556 (No connection could be made because the target machine actively refused it. - connect(2) for "localhost" port 17556)>
```

This is because v3.3.0 of this gem is downloading the latest version and ignoring the installed version of Edge.  The fix is to download the webdriver version that matches the installed version of Edge. I'll submit a PR to fix this. 

**Workaround**: Download the binary manually and or upgrade your Edge version.

kapoorlakshya:
Submitted PR #15 to fix this.

titusfortner:
PR #15 merged

PR #15, opened 2018-05-03
MSWebdriver now fetches the binary corresponding to the currently installed Edge version instead of the latest one.

kapoorlakshya:
This PR fixes issue #14 and updates the README.

Since `MicrosoftWebDriver.exe` is not backwards compatible, we cannot always update to the latest version. This PR allows `webdrivers` gem to download the correct version for the currently installed version of Edge.

It also fixes the bug where the gem would re-download the binary when `current != released`. It now assumes that an existing binary is the correct version and skips the download process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ in your Gemfile:
 
 in your project:
 
-`require 'webdrivers'
+`require 'webdrivers'`
 
 If there is a proxy between you and the Internet then you will need to configure
 the gem to use the proxy.  You can do this by calling the `configure` method.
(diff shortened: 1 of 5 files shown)


titusfortner:
Thanks for putting this together. Since we have subclasses for each driver, we don't want to add driver based conditionals to the superclass. What I'm thinking is that we can create `#correct_binary?` method and pull out:
```
current == released && File.exist?(binary)
```
Then `MSWebdriver#correct_binary?` would be:
```
File.exist?(binary)
```
Same with `download`. Subclass it for `MSWebdriver` and just change default:
```
def download(version = current)
  super
end
```

kapoorlakshya:
> Since we have subclasses for each driver, we don't want to add driver based conditionals to the superclass.

Good point, I definitely agree.  Will push the changes tonight.

kapoorlakshya:
@titusfortner I have updated the PR based on your suggestions. I have also fixed a failing `geckodriver` spec and all specs pass locally on Windows 10.

```
$ rspec
..........................

Finished in 29.82 seconds (files took 0.38402 seconds to load)
26 examples, 0 failures
```

However, the last `MSWebdriver` spec is failing on Travis because `current` relies on Powershell. Let me know if you want me to remove/modify this spec and what you think of the new changes.

titusfortner:
Excellent, thanks! I just released 3.3.1

PR #17, opened 2018-08-14
Use GitHub releases API instead of scraping the HTML

shepmaster:
Additionally, change to a version numbering sort that's more
correct (but still not perfect).

### DIFF
--- a/lib/webdrivers/geckodriver.rb
+++ b/lib/webdrivers/geckodriver.rb
@@ -15,7 +15,7 @@ module Webdrivers
       private
 
       def normalize(string)
-        string.match(/(\d+)\.(\d+\.\d+)/).to_a.map {|v| v.tr('.', '')}[1..-1].join('.').to_f
+        string.scan(/\d+/).map(&:to_i)
       end
 
       def downloads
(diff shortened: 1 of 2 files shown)


shepmaster:
Yes, the HTML changed again.

titusfortner:
Thanks for putting this together so quickly. I saw the Watir tests failing, so I threw together a fix for the html, it's released in 3.3.3.

The main problem with this implementation is the Github API [rate limiting](https://developer.github.com/v3/#rate-limiting). Since each new session makes an API call, 60/hr is not sufficient for large scale uses. Even with authentication, 5000/hr for *everyone* using the gem might not be enough... Maybe we can set it as the default, with a fall-back option to the existing implementation?

Also, I should update Geckodriver (well, all the drivers) to use `Gem::Version` instead of `Float`, like we did for [IEDriver](https://api.github.com/repos/mozilla/geckodriver/releases).

titusfortner:
Thanks for this PR. I'm going to close it since I don't see a way to get past the rate limiting of the API. If it becomes an issue we can investigate this again.

PR #21, opened 2018-09-18
fixed error with invalid regex

garyjob:
The invalid regex 
```
string.match(/ChromeDriver (\d\.\d+)/)[1]
```
was causing the following error 
```
NoMethodError - undefined method `[]' for nil:NilClass
```
at the following line of code
```
gems/webdrivers-3.2.2/lib/webdrivers/chromedriver.rb:12:in `current'
```

The actual reproduction in memory
```
string = "ChromeDriver 70.0.3538.16 (16ed95b41bb05e565b11fb66ac33c660b721f778)\n"
```

The modified regex below will support for when versions of the chrome driver has more than 1 digit in its first section.
```
string.match(/ChromeDriver (\d+\.\d+)/)[1]
```

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -9,7 +9,7 @@ module Webdrivers
         return nil unless downloaded?
         string = %x(#{binary} --version)
         Webdrivers.logger.debug "Current version of #{binary} is #{string}"
-        normalize string.match(/ChromeDriver (\d\.\d+)/)[1]
+        normalize string.match(/ChromeDriver (\d+\.\d+)/)[1]
       end
 
       def latest


allantokuda-zipnosis:
I'm seeing the same issue, and this change fixes it for me as well.

shepmaster:
@titusfortner could this and/or #20 get reviewed? Anything people can do to help?

titusfortner:
@shepmaster 3.4.1 should be working just fine.
I should merge #20 to use latest_release, though. I forget why I moved away from that, I think it went a while without being updated a few times. Still probably the better option going forward.

shepmaster:
So should this issue be closed?

titusfortner:
@shepmaster thanks for the ping on this. Merged #20 & released as 3.4.2

PR #22, opened 2018-10-19
Parse geckodriver versions page as HTML to fix JRuby usage

twalpole:


titusfortner:
@twalpole just released this with 3.4.3

PR #26, opened 2018-11-27
Use HTTPS to download Chromedriver and IEDriver

rbu:
Current webdrivers uses insecure transport communication to download code and runs no additional authenticity checks, allowing attackers with network access to execute arbitrary code (via an MITM attack, when the victim subsequently executes a webdriver).

### DIFF
--- a/spec/iedriver_spec.rb
+++ b/spec/iedriver_spec.rb
@@ -6,7 +6,7 @@ describe Webdrivers::IEdriver do
 
   it 'finds latest version' do
     old_version = Gem::Version.new("3.12.0")
-    future_version = Gem::Version.new("3.60.0")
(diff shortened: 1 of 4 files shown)


spec/iedriver_spec.rb
```diff
@@ -6,7 +6,7 @@
 
   it 'finds latest version' do
     old_version = Gem::Version.new("3.12.0")
-    future_version = Gem::Version.new("3.60.0")
+    future_version = Gem::Version.new("3.160.0")
```
rbu, line 9: Selenium released "141" last month, breaking this test.

titusfortner:
Merged: https://github.com/titusfortner/webdrivers/commit/645308c1fd986919e03e413a8cbe78df6218cd23

Issue #31, opened 2019-03-14
Checking latest Chromedriver version via LATEST_RELEASE is deprecated


jkotests:
From https://github.com/flavorjones/chromedriver-helper/issues/79, using Chromedriver's LATEST_RELEASE page for the latest version is deprecated.

We currently get 2.46, even though the latest is now 73.0.3683.68. Both support Chrome v73, so not a big deal. However, when Chrome v74 is released, we will presumably run into version mismatch errors (ie we'd still be downloading 2.46).

We need to update the process as documented in http://chromedriver.chromium.org/downloads/version-selection.

kapoorlakshya:
Hi @jkotests, I have submitted PR #32 if you'd like to review it when you get a chance.

kapoorlakshya:
Closed by #32.

PR #32, opened 2019-03-15
Chromedriver#latest_version now returns latest binary version corresponding to the installed Chrome version.

kapoorlakshya:
#31 - Google now matches the `chromedriver` version to Chrome browser version ([source](http://chromedriver.chromium.org/downloads/version-selection)). This PR updates `Chromedriver#latest_version` to download the latest binary corresponding to the installed Chrome (major) version.

Without this change, the gem fetches `chromedriver` version 2.46 as the latest one regardless of your Chrome version.

**Example**

If you have Chrome 72.0.3626.121 installed, the gem will now fetch the latest binary release for version 72.0, which is 72.0.3626.69.

**Specs**

Windows - Pass (local @ Windows 10)
Linux - Pass (Travis)
Mac - Pass (Travis)

Updated `travis.yml` to install Chrome, and to run specs on Linux and Mac..

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -14,7 +14,11 @@ module Webdrivers
 
       def latest_version
         raise StandardError, "Can not reach site" unless site_available?
-        Gem::Version.new(get(URI.join(base_url, "LATEST_RELEASE")))
+
+        # Latest webdriver release for installed Chrome version
+        release_file     = "LATEST_RELEASE_#{release_version}"
+        latest_available = get(URI.join(base_url, release_file))
+        Gem::Version.new(latest_available)
       end
 
       private
@@ -31,11 +35,11 @@ module Webdrivers
         Webdrivers.logger.debug "Versions previously located on downloads site: #{@downloads.keys}" if @downloads
 
         @downloads ||= begin
-          doc = Nokogiri::XML.parse(get(base_url))
+          doc   = Nokogiri::XML.parse(get(base_url))
           items = doc.css("Contents Key").collect(&:text)
-          items.select! {|item| item.include?(platform)}
+          items.select! { |item| item.include?(platform) }
           ds = items.each_with_object({}) do |item, hash|
-            key = normalize item[/^[^\/]+/]
+            key       = normalize item[/^[^\/]+/]
             hash[key] = "#{base_url}/#{item}"
           end
           Webdrivers.logger.debug "Versions now located on downloads site: #{ds.keys}"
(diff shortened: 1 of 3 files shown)


kapoorlakshya:
Alright! Specs now pass on all three - Windows, Linux, and OSX.

lib/webdrivers/chromedriver.rb
```diff
@@ -43,6 +47,42 @@ def downloads
         end
       end
 
+      # Returns major version from the currently installed Chrome version
+      def chrome_major_version
+        chrome_version[0..1]
```
jkotests: You might want to grab everything up to the first dot instead of just the first 2 characters. Otherwise, you'll have to update this in a year or two when version 100 is released.

lib/webdrivers/chromedriver.rb
```diff
@@ -14,7 +14,11 @@ def current_version
 
       def latest_version
         raise StandardError, "Can not reach site" unless site_available?
-        Gem::Version.new(get(URI.join(base_url, "LATEST_RELEASE")))
+
+        # Latest webdriver release for installed Chrome version
+        release_file     = "LATEST_RELEASE_#{chrome_major_version}"
```
jkotests: Not sure if it matters, but Chromedriver's suggestion is to use the first three numbers of the version:

> Take the Chrome version number, remove the last part, and append the result to URL "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_". For example, with Chrome version 72.0.3626.81, you'd get a URL "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_72.0.3626".

Just using the major version was only mentioned if using a non-stable/beta Chrome release. Again, I'm not sure how much it matters.

lib/webdrivers/chromedriver.rb
kapoorlakshya: Good point! The PR has been updated.

lib/webdrivers/chromedriver.rb
kapoorlakshya: Thanks for pointing that out. I have renamed `#chrome_major_version` to `#release_version` - it uses the first three version numbers now.

titusfortner:
Ugh. This is another example of me walking through my code wondering "who the hell wrote this crap?"

lib/webdrivers/chromedriver.rb
```diff
@@ -31,18 +35,57 @@ def downloads
...
           ds
         end
       end
 
+      # Returns release version from the currently installed Chrome version
+      #
+      # @example
+      #   73.0.3683.75 -> 73.0.3683
+      def release_version
+        chrome_version.scan(/\d+\.\d+\.\d+/)[0]
```
titusfortner: We should avoid scan if we don't need the Array: `chrome_version[/\d+\.\d+\.\d+\.\d+/]`
Not sure if rubocop would flag this (https://github.com/rubocop-hq/ruby-style-guide#regexp-string-index)

lib/webdrivers/chromedriver.rb
```diff
@@ -31,18 +35,57 @@ def downloads
...
+                when /linux/
+                  chrome_on_linux
+                when 'mac'
+                  chrome_on_mac
+                else
+                  raise NotImplementedError, 'Your OS is not supported by webdrivers gem.'
+              end.chomp
+
+        # Google Chrome 73.0.3683.75 -> 73.0.3683.75
+        ver.scan(/\d+\.\d+\.\d+\.\d+/)[0]
```
titusfortner: We can do: `ver[/(\d|\.)+/]`

lib/webdrivers/chromedriver.rb
```diff
@@ -31,18 +35,57 @@ def downloads
...
+        ps  = "(Get-Item (Get-ItemProperty '#{reg}').'(Default)').VersionInfo.ProductVersion"
+        `powershell #{ps}`
+      end
+
+      def chrome_on_linux
+        `google-chrome --version`
+      end
+
+      def chrome_on_mac
+        loc = Shellwords.escape '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
```
titusfortner, line 86: Oh cool, I've learned about a new useful class, thanks.

titusfortner (changes requested):
This is good to start. Obviously there are several things that can be fixed now (`#downloads` should be unnecessary now). We can tackle those separately, though.

lib/webdrivers/chromedriver.rb
kapoorlakshya: Oh nice! I didn't know we could do that. Updated to `chrome_version[/\d+\.\d+\.\d+/]` to get the first 3 versions.

lib/webdrivers/chromedriver.rb
kapoorlakshya: Done!

lib/webdrivers/chromedriver.rb
kapoorlakshya: Yea, pretty useful. I found it in Sean Linsley's gist here: https://gist.github.com/seanlinsley/36a204862f08942b0667e941331f3a75

PR #39, opened 2019-03-25
Chromedriver now checks Selenium::WebDriver::Chrome#path.

### LINKED ISSUE #38: [3.7.0] Doesn't work with chromium on Linux
I guess the issue is [here](https://github.com/titusfortner/webdrivers/blob/ee698cb3c41ee39f5ee13f8d32cbd4623f80e7a6/lib/webdrivers/chromedriver.rb#L82). This checks for `google-chrome` while I, for example, use `chromium`. The code should be made to handle that scenario.

```
Errno::ENOENT:
  No such file or directory - google-chrome
# ./features/build/tmp/jrubyExec/gems/webdrivers-3.7.0/lib/webdrivers/chromedriver.rb:82:in `chrome_on_linux'
# ./features/build/tmp/jrubyExec/gems/webdrivers-3.7.0/lib/webdrivers/chromedriver.rb:64:in `chrome_version'
# ./features/build/tmp/jrubyExec/gems/webdrivers-3.7.0/lib/webdrivers/chromedriver.rb:55:in `release_version'
# ./features/build/tmp/jrubyExec/gems/webdrivers-3.7.0/lib/webdrivers/chromedriver.rb:19:in `latest_version'
# ./features/build/tmp/jrubyExec/gems/webdrivers-3.7.0/lib/webdrivers/common.rb:37:in `desired_version'
# ./features/build/tmp/jrubyExec/gems/webdrivers-3.7.0/lib/webdrivers/common.rb:16:in `update'
# ./features/build/tmp/jrubyExec/gems/webdrivers-3.7.0/lib/webdrivers/selenium.rb:7:in `driver_path'
# ./features/build/tmp/jrubyExec/gems/selenium-webdriver-3.141.0/lib/selenium/webdriver/chrome/driver.rb:39:in `initialize'
# ./features/build/tmp/jrubyExec/gems/selenium-webdriver-3.141.0/lib/selenium/webdriver/common/driver.rb:44:in `for'
# ./features/build/tmp/jrubyExec/gems/selenium-webdriver-3.141.0/lib/selenium/webdriver.rb:86:in `for'
# ./features/build/tmp/jrubyExec/gems/capybara-3.15.0/lib/capybara/selenium/driver.rb:32:in `browser'
# ./features/build/tmp/jrubyExec/gems/capybara-3.15.0/lib/capybara/selenium/driver.rb:139:in `current_window_handle'
# ./features/build/tmp/jrubyExec/gems/capybara-3.15.0/lib/capybara/session.rb:438:in `current_window'
# ./features/spec_helper.rb:89:in `block in <main>'
```

kapoorlakshya:
- Chromedriver now checks `Selenium::WebDriver::Chrome#path` before defaulting to Google Chrome. Addresses #38.
- Now downloads `chromedriver` v2.46 if Chrome/Chromium version is less than 70. `LATEST_FILE_*` only available for versions > 70.
- `Chromedriver#current_version` now returns full version - `73.0.3683.68` instead of `73.0`.

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -5,19 +5,26 @@ module Webdrivers
     class << self
 
       def current_version
-        Webdrivers.logger.debug "Checking current version"
+        Webdrivers.logger.debug 'Checking current version'
         return nil unless downloaded?
-        string = %x(#{binary} --version)
-        Webdrivers.logger.debug "Current version of #{binary} is #{string}"
-        normalize string.match(/ChromeDriver (\d+\.\d+)/)[1]
+
+        ver = `#{binary} --version`
+        Webdrivers.logger.debug "Current #{binary} version: #{ver}"
+
+        # Matches 2.46, 2.46.628411 and 73.0.3683.75
+        normalize ver[/\d+\.\d+(\.\d+)?(\.\d+)?/]
       end
 
       def latest_version
-        raise StandardError, "Can not reach site" unless site_available?
+        raise StandardError, 'Can not reach site' unless site_available?
+
+        # Versions before 70 do not have a LATEST_RELEASE file
+        return Gem::Version.new('2.46') if release_version < '70.0.3538'
 
         # Latest webdriver release for installed Chrome version
         release_file     = "LATEST_RELEASE_#{release_version}"
         latest_available = get(URI.join(base_url, release_file))
+        Webdrivers.logger.debug "Latest version available: #{latest_available}"
         Gem::Version.new(latest_available)
       end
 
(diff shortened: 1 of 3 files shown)


lib/webdrivers/chromedriver.rb
```diff
@@ -73,18 +80,56 @@ def chrome_version
...
+      # Returns true if user defines path to the browser executable.
+      #
+      def browser_defined?
+        !browser_exe.nil?
+      end
+
+      #
+      # Returns user defined browser executable path from Selenium::WebDrivers::Chrome#path.
+      #
+      def browser_exe
```
titusfortner: Let's use `browser_binary` since it isn't an exe except on Windows

lib/webdrivers/chromedriver.rb
```diff
@@ -73,18 +80,56 @@ def chrome_version
...
+        # Default to Google Chrome
+        executable = Shellwords.escape '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
+        Webdrivers.logger.debug "Browser executable: #{executable}"
+        `#{executable} --version`.strip
+      end
+
+      #
+      # Returns true if user defines path to the browser executable.
+      #
+      def browser_defined?
```
titusfortner: I think we should just use `!browser_binary` instead of defining a new helper method for this? I think the method is clear enough by itself without making the class longer?

lib/webdrivers/chromedriver.rb
```diff
@@ -73,18 +80,56 @@ def chrome_version
...
+        Webdrivers.logger.debug "Browser executable: '#{executable}'"
+        `#{executable} --product-version`.strip
       end
 
       def chrome_on_mac
-        loc = Shellwords.escape '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
-        `#{loc} --version`
+        if browser_defined?
+          Webdrivers.logger.debug "Browser executable: '#{browser_exe}'"
+          return `#{browser_exe} --version`.strip if browser_defined?
```
titusfortner: When is browser defined on line 111 if it isn't defined on 109?

spec/chromedriver_spec.rb
```diff
@@ -25,29 +25,45 @@
...
+    latest_ver = chromedriver.latest_version.version
     expect(cur_ver).to eq latest_ver
   end
 
   it 'downloads specified version by Float' do
     chromedriver.remove
     chromedriver.version = 2.29
     chromedriver.download
-    expect(chromedriver.current_version.version).to eq '2.29'
+    expect(chromedriver.current_version.version[/\d+.\d+/]).to eq '2.29'
```
titusfortner: let's use `to include` rather than Regexp here

lib/webdrivers/chromedriver.rb
```diff
@@ -73,18 +80,56 @@ def chrome_version
...
-        reg = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\chrome.exe'
-        ps  = "(Get-Item (Get-ItemProperty '#{reg}').'(Default)').VersionInfo.ProductVersion"
-        `powershell #{ps}`
+        if browser_defined?
+          Webdrivers.logger.debug "Browser executable: '#{browser_exe}'"
+          return `powershell (Get-ItemProperty '#{browser_exe}').VersionInfo.ProductVersion`
+        end
+
+        # Default to Google Chrome
+        reg        = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\chrome.exe'
```
titusfortner, line 93: It looks like this is how chromedriver is finding the values: https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver#requirements
I'm not sure if the above is different, but it might be different depending on the version of Windows.

titusfortner (changes requested):
Good work on this, I have a few suggestions here.

lib/webdrivers/chromedriver.rb
kapoorlakshya: Sure. I was using "exe" as in executable, but `browser_binary` sounds less confusing.

lib/webdrivers/chromedriver.rb
kapoorlakshya: Makes sense. Updated!

lib/webdrivers/chromedriver.rb
kapoorlakshya: Ah, I went from inline to block style and forgot to remove that.

lib/webdrivers/chromedriver.rb
```diff
@@ -73,18 +80,56 @@ def chrome_version
...
+        Webdrivers.logger.debug "Browser executable: '#{executable}'"
+        ps = "(Get-Item (Get-ItemProperty '#{reg}').'(default)').VersionInfo.ProductVersion"
+        `powershell #{ps}`.strip
       end
 
       def chrome_on_linux
-        `google-chrome --version`
+        if browser_defined?
+          Webdrivers.logger.debug "Browser executable: '#{browser_exe}'"
+          return `#{browser_exe} --product-version`.strip if browser_defined?
```
beauraF:
```suggestion
          return `#{browser_exe} --product-version`.strip
```

lib/webdrivers/chromedriver.rb
kapoorlakshya: Yea, it's different. I tested the registry and it works on Windows 7, 10, Windows Server 2008, and 2016. Not sure if we want to support XP and 7 still.

titusfortner: Yeah, windows 7 is not EOL, yet.

kapoorlakshya: Yeah, I meant to say XP only. Registry works on Windows 7+.

titusfortner (approved):
Let's merge it. I've asked for clarification from lead chromedriver dev. We can update the code when we hear back if we have different defaults.

PR #40, opened 2019-03-25
add support for heroku/heroku-buildpack-google-chrome

beauraF:
Hi,

On heroku CI, with [heroku/heroku-buildpack-google-chrome](https://github.com/heroku/heroku-buildpack-google-chrome), `webdrivers` having some issues.

Heroku replaces the `google-chrome` command with the following script:
```bash
exec $HOME/.apt/opt/google/chrome/chrome --headless --no-sandbox --disable-gpu --remote-debugging-port=9222 $@
```

This script does not support the `--version` option. In fact, when you make a call to: 
```ruby
def chrome_on_linux
  `google-chrome --version`
end
```

Instead of getting the version, the script runs Google Chrome headlessly in forground.

I propose to add the support of this buildpack by checking the presence of the `shims` provided for this purpose by heroku. : https://github.com/heroku/heroku-buildpack-google-chrome#shims-and-command-line-flags

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -79,7 +79,7 @@ module Webdrivers
       end
 
       def chrome_on_linux
-        `google-chrome --version`
+        `#{ENV['GOOGLE_CHROME_BIN'] || 'google-chrome'} --version`
       end
 
       def chrome_on_mac


kapoorlakshya:
Hi @beauraF, thanks for the PR. Are you using Selenium with `heroku-buildpack-google-chrome`?

You're right about `--version` launching the browser in the foreground. It should be `--product-version`.  I've fixed it in #39 along with support for custom browser executable path or shims when using Selenium.

beauraF:
Hi @kapoorlakshya, thanks for ur answer.

Yes, we are using `heroku-buildpack-google-chrome` on Heroku CI with `selenium-webdriver`.
I tried ur fix on our CI and unfortunately it doesn't work. Same issue.

kapoorlakshya:
Ah okay. Sorry, I'm not too familiar with Heroku, so let me ask you this - 

1. Is it possible to set `Selenium::WebDriver::Chrome.path` to the actual `google-chrome` binary that the 
 shim wraps around?
2. Does `which google-chrome` return `$HOME/.apt/opt/google/chrome/chrome` or `/usr/bin/google-chrome`?
3. Which path does `ENV['GOOGLE_CHROME_BIN']` point to?

Just checking to see if we can make this work with the current code in #39. If we cannot, I'd be happy to implement support for this.

beauraF:
No problem. :)

> Is it possible to set Selenium::WebDriver::Chrome.path to the actual google-chrome binary that the shim wraps around?

I'm trying to find a suitable way to do it with our current stack.

> Does which google-chrome return $HOME/.apt/opt/google/chrome/chrome or /usr/bin/google-chrome?
```
~ $ which google-chrome
/app/.apt/usr/bin/google-chrome
```

> Which path does ENV['GOOGLE_CHROME_BIN'] point to?
```
~ $ echo $GOOGLE_CHROME_BIN
/app/.apt/opt/google/chrome/chrome
```

beauraF:
It's okay. By using #39 and setting `Selenium::WebDriver::Chrome.path`, everything is green.
I close this PR. Thank u @kapoorlakshya.

Just, do you know the difference between [Selenium::WebDriver::Chrome.path](https://seleniumhq.github.io/selenium/docs/api/rb/Selenium/WebDriver/Chrome.html#path-class_method) and [Selenium::WebDriver::Chrome::Options.binary](https://seleniumhq.github.io/selenium/docs/api/rb/Selenium/WebDriver/Chrome/Options.html#binary-instance_method) ?

kapoorlakshya:
Hi @beauraF, I'm glad it works for you now.

> Just, do you know the difference between Selenium::WebDriver::Chrome.path and Selenium::WebDriver::Chrome::Options.binary ?

It sounds like they technically do the same thing - allow us to set path to the chrome binary. I found some info on `binary` here - http://chromedriver.chromium.org/capabilities. See the the `chromeOptions` section.

@titusfortner - Any insight on this?

titusfortner:
`Selenium::WebDriver::Chrome.path` only applies locally. What gets passed in to the Options is supposed to work both locally and remotely. 

The current Selenium code is actually setting `Selenium::WebDriver::Chrome::Options.binary` to `Selenium::WebDriver::Chrome.path` if it is specified. This is "wrong" because it will cause problems when executing on a remote server. We're still trying to hammer out the right way to fix it.

kapoorlakshya:
The latest version of the google-chrome buildpack no longer needs the workaround described [here](https://github.com/titusfortner/webdrivers/wiki/Heroku-buildpack-google-chrome). We've updated the SHIM to support `webdrivers`. See https://github.com/heroku/heroku-buildpack-google-chrome/pull/73/ for more information.

beauraF:
Insane. Thank you very much @kapoorlakshya !

PR #43, opened 2019-03-26
Fix chromium version handling

joe81:
Fix the handling of chrome version with additional version informations like "Using PPAPI flash.\n73.0.3683.75".

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -80,7 +80,7 @@ module Webdrivers
         end
 
         # Google Chrome 73.0.3683.75 -> 73.0.3683.75
-        ver[/(\d|\.)+/]
+        ver[/\d+\.\d+\.\d+\.\d+/]
       end
 
       def chrome_on_windows


kapoorlakshya:
Hi @joe81, thanks for your PR! I would've never caught that since Flash is disabled on my browsers now 🙂.

Your change looks good, and it makes me think that maybe we should use something more along the lines of  `\d+\.\d+\.\d+\.\d+` (https://rubular.com/r/QSFR4enA1Wgq2N).  The Regex in your PR will work as long as the version is at the end of the string. However, if Google decides to add some other info after it, such as "x64" or "win", it'll cease to work. Just thinking ahead to keep it low maintenance. 

Let me know what you think.

joe81:
Hello @kapoorlakshya , thank you for your quick answer. Your regex is much more robust than the previous one. From my point of view you should use it. I would close my PR because it is obsolete. How do you see that? When could you make the change available?

kapoorlakshya:
@joe81 Feel free to update your PR with the preferred regex, and then squash the commits. I'll approve and merge it it once the specs pass on Travis. Hoping to do a v3.7.2 release very soon..

joe81:
@kapoorlakshya I've adjusted the PR.

kapoorlakshya:
Thank you!

PR #47, opened 2019-03-26
Rubocop

stephengroat:
closes #35

PR #54, opened 2019-03-29
Rubocop (with rspec) and code coverage

stephengroat:
fix of #47 

cc @kapoorlakshya https://github.com/titusfortner/webdrivers/pull/47#issuecomment-477714310


kapoorlakshya:
@stephengroat Thank you again!

PR #60, opened 2019-04-05
Remove dependency on `net_http_ssl_fix`. Update README. Fixes #59.

### LINKED ISSUE #59: Don't depend on net_http_ssl_fix
This should be optional, not required. I suggest adding a note in the README for users who want to use it. Otherwise, I see this as a major security risk.

ioquatix:


kapoorlakshya:
TODO for me:

- [x] Update [wiki](https://github.com/titusfortner/webdrivers/wiki/SSL-Certificate-Verify-Failed) entry when this gets merged.

kapoorlakshya:
@ioquatix Can you please squash your commits locally and force push? Also, thanks for updating the heading sizes in the README. They look fine on my 4K monitor, but look huge when I now see it at 1080p or lower.

ioquatix:
I was going to do a separate PR for the README but in the end it was such a small change.

The usage of nested `#` is pretty standard for markdown, and maps to `<h1>`...`<h6>` depending on the number of `#`. It shouldn't look any different on different resolutions excepting how your OS handles scaling of fonts.

ioquatix:
> Can you please squash your commits locally and force push.

You can do this when you choose "Squash & Merge" from the merge options. GitHub will do it automatically.

kapoorlakshya:
> You can do this when you choose "Squash & Merge" from the merge options. GitHub will do it automatically.

You're right. I keep forgetting about it.

titusfortner:
Oh, I'm blind and didn't see the PR, just the issue. Completely agree this is better.

Do we want to go further and tell users in the README that while it is probably the easiest solution, it might not be the best solution?

titusfortner (approved):

kapoorlakshya:
> Do we want to go further and tell users in the README that while it is probably the easiest solution, it might not be the best solution?

Yeah, may as well, and include a link to this page for more info - https://guides.rubygems.org/ssl-certificate-update/

ioquatix:
I'll leave this in your hands.

PR #68, opened 2019-04-17
Webdrivers.net_http_ssl_fix now raises an exception.

kapoorlakshya:
The exception informs the user this fix is no longer available, and points them to https://github.com/titusfortner/webdrivers#ssl_connect-errors.

Also added a link to the RubyGems [guide](https://guides.rubygems.org/ssl-certificate-update/) for SSL certificate related fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ Add the following to your code:
 require 'net_http_ssl_fix'
 ````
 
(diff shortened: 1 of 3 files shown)


kapoorlakshya:
@titusfortner I'm merging this to get v3.8.0 out. If you would like any changes in here, we can do that in a point release.

PR #76, opened 2019-04-27
Add config to disable automatic updates.

kapoorlakshya:
* Add config `Webdrivers.auto_update` to enable/disable automatic updates.
* When set to `true`, which is the default setting, it will always perform an update check before passing the driver to Selenium.
* When set to `false` it disables the update check as long as a driver binary exists. 
* If `Webdrivers.auto_update` is set to `false` and a binary does not exist, it will download the driver once and never check for updates as long as the updates are disabled. This is useful when you do want to update the driver without altering the config in your project - simply delete the existing binary to trigger a one-time update.

Other changes:

* Added network request counter for testing purpose.
* Current Chrome/Chromium version is now printed to the debug log. I forgot to add this in my PR for #38.
* Disabled `Metrics/PerceivedComplexity` in rubocop because this features increases the complexity in `Common#update` from 7 to 9. We can deal with this in the rewrite for v4.
* Added `ffi` as a development dependency. `selenium-webdriver` -> `childprocess` requires it on Windows, but does not list it as a dependency.

**EDIT**: I am now thinking that addition of `ffi` should be its own PR so its easy to track from the commits. Let me know if you have a preference.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,8 @@ Metrics/AbcSize:
   Enabled: false
 Metrics/CyclomaticComplexity:
   Enabled: false
+Metrics/PerceivedComplexity:
+  Enabled: false
 Style/Documentation:
   Enabled: false
 RSpec/ExampleLength:
--- a/README.md
+++ b/README.md
@@ -62,6 +62,22 @@ You can also trigger the update in your code, but it is not required:
 Webdrivers::Chromedriver.update
 ```
 
+### Automatic Updates
+
+Automatic updates are enabled by default, which means `webdrivers` will 
+always check for updates before passing the driver to Selenium. 
+If you do not want this, you can disable automatic updates in your 
+project:
+
+```ruby
+Webdrivers.auto_update = false
+```
+
+If a driver binary does not already exist, the gem will download it
+once and then never check for updates as long you have the updates disabled. 
+When you do want to download an updated driver without altering this 
+setting, simply delete the existing driver binary.
+
 ### Proxy
 
 If there is a proxy between you and the Internet then you will need to configure
--- a/lib/webdrivers.rb
+++ b/lib/webdrivers.rb
@@ -8,7 +8,8 @@ require 'webdrivers/mswebdriver'
 
 module Webdrivers
   class << self
-    attr_accessor :proxy_addr, :proxy_port, :proxy_user, :proxy_pass, :install_dir
+    attr_accessor :proxy_addr, :proxy_port, :proxy_user,
+                  :proxy_pass, :install_dir, :auto_update
 
     def logger
       @logger ||= Webdrivers::Logger.new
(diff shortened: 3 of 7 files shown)


januszm:
Looks great! Thanks. And true, it'd be better to add `ffi` in separate PR for debugging purposes (e.g. with `git bisect`).

🎉

titusfortner:
So what is the desired behavior for the driver and the browser having a mismatch (for Chrome)?

I'm thinking we should require users to be explicit about what version of driver they want to use and throw a warning that there is a mismatch, rather than just turning on a toggle not to update...

januszm:
@titusfortner  I think that this question deserves its own Github Issue and Pull Request.

titusfortner:
@januszm I'm saying that I don't think we should merge this PR because it is not the desired behavior. Setting the version is currently implemented and won't need a PR. (figuring out what warning where will need to be done elsewhere)

kapoorlakshya:
> So what is the desired behavior for the driver and the browser having a mismatch (for Chrome)?

@titusfortner I forgot to mention this earlier - The idea behind this config is to match the behavior of `chromedriver-helper`. The user is responsible for updating the driver.

If there is a mismatch (Selenium error), they can simply delete the existing driver which will force a download (update) once. 

When `auto_update` is set to true, which is the default, `webdrivers` will continue to work as it has.

> Setting the version is currently implemented

`#update` still makes a few network calls (See debug log in #80) because of the `site_available?` check. I know we'll fix this soon, but multiple users coming from `chromedriver-helper` (I assume) have asked for no network calls at all. It may be nice to have this feature roll out in a pre-v4 release so users don't have to wait for a while.

**EDIT**: Alternatively, I'd prefer to work on a better/preferred solution with you now that you are done with SauceCon.

titusfortner:
Selenium won't throw an error if the driver is set to 72.x and the browser is 74.x. Selenium doesn't care.

The main reason I made a separate gem instead of using chromedriver-helper is because I wanted it to automatically download by default. If a user does not want to download by default (locked browser to not auto-update, or using a specific browser version with a CI tool), then they need to be responsible for specifying the driver version they want to use.

As for the network calls, yes, we need to refactor. It should look something approximating this:

1. If driver version specified and that driver is present --> no network calls
2. If driver version is specified and that driver is not present --> one network call to download
3. If driver version is not specified then one network call to check & one to download (if necessary)

kapoorlakshya:
Fair enough :)

As discussed on Slack, you'll have some time work on these issues this week, so I'll go ahead and close this and #49.

januszm:
@titusfortner there's one more scenario:

4. The user does not care about the driver version, it must be modern enough to run simple system / feature tests (including Javascript) in Rails.

I would vote for the re-opening and merging of this PR because it solves the common problem and it does not change the default behaviour (auto-updates). There must be a way to disable automatic updates when running tests, preceded by a driver installation using the CI / CD configuration.

titusfortner:
Except when the user doesn't care about the version, things break when the browser updates, which is the primary point of this gem. 

> There must be a way to disable automatic updates when running tests, preceded by a driver installation using the CI / CD configuration.

Determining the version of browser you are using and specifying the associated driver version you want in your code is that way.

PR #81, opened 2019-04-30
download chromedriver directly instead of parsing the index

titusfortner:
We shouldn't need to parse the download page when we know the exact file we want from chrome.

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -34,24 +34,30 @@ module Webdrivers
 
       private
 
+      def platform
+        if Selenium::WebDriver::Platform.linux?
+          'linux64'
+        elsif Selenium::WebDriver::Platform.mac?
+          'mac64'
+        else
+          'win32'
+        end
+      end
+
       def file_name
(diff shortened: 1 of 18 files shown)


lib/webdrivers/chromedriver.rb
```diff
@@ -30,28 +30,30 @@ def latest_version
 
       private
 
+      def platform
+        if Selenium::WebDriver::Platform.linux?
```
kapoorlakshya, line 38: How about we return `Selenium::WebDriver::Platform` here and then call `#linux?`, `#mac?`, and `#linux?` on it instead of comparing strings in `#file_name` and `#chrome_version`?

```ruby
platform.windows?
```

instead of

```ruby
platform == 'win32'
```

kapoorlakshya: This will also fix the rubocop offense on line 35.

kapoorlakshya (commented):
Please see if `#platform` can be improved.

lib/webdrivers/chromedriver.rb
titusfortner: we can just call `#windows?`, and it's definitely the more precise implementation for this. Updated.

PR #82, opened 2019-04-30
remove check for site available 

titusfortner:
I'm not sure if `SocketError` is everything we need to rescue to make the error message good, but this should remove an unnecessary network request.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,19 +1,31 @@
 require: rubocop-rspec
+
 Metrics/LineLength:
-  Enabled: false
+  Max: 120
+
 Metrics/MethodLength:
   Enabled: false
+
 Metrics/BlockLength:
-  Enabled: false
+  Exclude:
+    - 'spec/**/*'
+
 Metrics/ClassLength:
-  Enabled: false
+  Exclude:
+    - 'lib/webdrivers/common.rb'
+
 Metrics/AbcSize:
-  Enabled: false
+  Exclude:
+    - 'lib/webdrivers/common.rb'
+
 Metrics/CyclomaticComplexity:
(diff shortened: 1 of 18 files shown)


kapoorlakshya:
Based on this [gem](https://github.com/edward/net_http_exception_fix/blob/master/lib/net_http_exception_fix.rb), we can rescue from a few of these common errors:

```ruby
Timeout::Error
Errno::ETIMEDOUT
Errno::EINVAL
Errno::ECONNRESET
Errno::ECONNREFUSED
EOFError
Net::HTTPBadResponse
Net::HTTPHeaderSyntaxError
Net::ProtocolError
```

titusfortner:
Socket error is what needs to get rescued since it's what you get when you turn off your internet and try to run. Everything else should return `response.value`, which may or may not be what we want since not everything is validating that the right thing happened after calling `#get`

@twalpole any recommendations for how far down this rabbit hole we need to go?

twalpole:
I usually try to look at it from the users perspective - If the user has set up webdrivers to run the latest version - and you can't get to the site that's an error because you can't guarantee the user they're running the latest.  If the user has locked the version, doesn't yet have a copy of that version, and the site isn't available that's also an error because you're not running the version the user wants.  I would ask why you're trying to rescue anything specific here - catch StandardError and raise an error with a nice message setting the initial exception as the cause .

kapoorlakshya (approved):
Specs pass on Windows as well. I'll leave it up to you on how we should handle the exceptions from `#get`.

titusfortner:
I'm jumping the gun on this, it needs a rework. I'm closing it for now, and will reopen when I have my logic figured out.

Issue #79, opened 2019-04-29
Improve error message when Driver for Chrome Dev (v75.x.x) is not found


EdgarOrtegaRamirez:
Hey! Just migrated from chrome-driver to webdrivers.
I have installed Chrome Dev Channel (75.0.3770.8 dev) in my Mac. This was the output when I ran tests without any additional configuration. It would be cool if this message could be improved when the webdriver version couldn't be found and it is falling back to the installed chrome app. The tests didn't work when this happened.

```sh
➜ bin/rails test:system
Run options: --seed 21855

# Running:

Capybara starting Puma...
* Version 3.12.1 , codename: Llamas in Pajamas
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:50633
2019-04-29 12:35:57 DEBUG Webdrivers Looking for Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:35:57 DEBUG Webdrivers Get response: #<Net::HTTPOK 200 OK readbody=true>
2019-04-29 12:35:57 DEBUG Webdrivers Found Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:35:57 DEBUG Webdrivers Looking for Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:35:58 DEBUG Webdrivers Get response: #<Net::HTTPOK 200 OK readbody=true>
2019-04-29 12:35:58 DEBUG Webdrivers Found Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:35:58 DEBUG Webdrivers Browser executable: /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome
2019-04-29 12:35:58 DEBUG Webdrivers Browser executable: /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome
2019-04-29 12:35:59 DEBUG Webdrivers Get response: #<Net::HTTPNotFound 404 Not Found readbody=true>
2019-04-29 12:35:59 DEBUG Webdrivers Looking for Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:35:59 DEBUG Webdrivers Get response: #<Net::HTTPOK 200 OK readbody=true>
2019-04-29 12:35:59 DEBUG Webdrivers Found Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:35:59 DEBUG Webdrivers Looking for Site: https://chromedriver.storage.googleapis.com
[...]

PR #83, opened 2019-05-01
specify behavior for unmatched drivers

titusfortner:
This is to address #79 

Options to address this are:
1. Raise exception if an associated driver is not found & ask the user to supply the desired version explicitly
2. Use the latest released version with a direct download
3. Parse the DOM of the release page to find the latest beta version and use that

This PR goes with option 1 because I think the likelihood of there being a problem with the mismatch with the other two options is high enough that I'd rather it not be the default and have users wonder why certain tests are failing. If you want to use the latest/greatest beta browser in your testing, you should be able to figure out how to specify the desired version in your code.

Let me know if there is a better way to address this, or if my arguments against failing more gracefully specifically in the case of Chrome aren't useful.

Implementation wise, this changes all of the references to a version to parse into `Gem::Version` instances, which is the only way to ensure comparisons are correct.

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -14,7 +14,7 @@ module Webdrivers
         Webdrivers.logger.debug "Current #{binary} version: #{ver}"
 
         # Matches 2.46, 2.46.628411 and 73.0.3683.75
-        normalize ver[/\d+\.\d+(\.\d+)?(\.\d+)?/]
+        normalize_version ver[/\d+\.\d+(\.\d+)?(\.\d+)?/]
       end
 
       def latest_version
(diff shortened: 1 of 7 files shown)


lib/webdrivers/chromedriver.rb
```diff
@@ -23,13 +23,24 @@ def latest_version
...
-        Webdrivers.logger.debug "Latest version available: #{latest_available}"
-        @latest_version = Gem::Version.new(latest_available)
+        release_file = "LATEST_RELEASE_#{release_version}"
+        begin
+          latest_applicable = normalize_version(get(URI.join(base_url, release_file)))
+        rescue Net::HTTPServerException
+          latest_release = normalize_version(get(URI.join(base_url, 'LATEST_RELEASE')))
+          Webdrivers.logger.debug "Unable to find a driver for: #{release_version}"
+
+          msg = release_version > latest_release ? 'you are using a non-production version of Chrome; ' : ''
```
twalpole: Maybe the message should be 'You appear to be using a non-production version...'

lib/webdrivers/chromedriver.rb
```diff
@@ -23,13 +23,24 @@ def latest_version
...
+          msg = "#{msg}please set `Chromedriver#version=` to an known chromedriver version: " \
```
twalpole: Probably should be 'set `Webdrivers::Chrome.version = <desired driver version>`' (the Webdrivers module namespace being the important change)

twalpole (commented):
Just a couple of comments on the messages, otherwise LGTM.  If the user is running a browser that webdrivers can't determine a compatible driver for it should definitely be up to the user to specify which one to use.

kapoorlakshya (approved):
Good call on normalizing all String based versions that I forgot in #32 :)

PR #84, opened 2019-05-01
Download to temp files to prevent threads/processes colliding

twalpole:


twalpole:
Not sure this is a full solution to the issue since multiple threads/processes would still overwrite each other on the final extracted binaries - but it should remove the current issue of collisions while downloading

twalpole:
A different solution may be to use a file lock when downloading to act as a cross process mutex and just block other processes while downloading/extracting

titusfortner:
Well, this is better than current code, so I guess we can start here and see where we need to go. 

We should also get Appveyor set up so that it is easier to validate that all of these things work properly on Windows.

Issue #36, opened 2019-03-19
Public Methods


titusfortner:
Let's make sure that browser versions and driver versions are properly and usefully publicly accessible

kapoorlakshya:
As discussed on Slack, these would be useful to have for all supported browsers:

```
#
# Returns installed browser version
#
Webdrivers.chrome.version 

#
# Returns current webdriver version
#
Webdrivers.chrome.driver_version 

#
# Returns ~/.webdrivers/chromedriver or user given path
#
Webdriver.chrome.driver_path
```

And maaaaybe.... `Webdrivers.chrome.latest_driver_version` to easily compare what you have vs latest available for download.

titusfortner:
What we have now:
`#update`
`#remove`
`#download` --> This can be made private, there's no reason to download independent of updating

`#version` (requested version)
`#latest_version` (network call to lookup)
`#desired_version` (either requested or latest) --> This can be made private, users don't need this
`#current_version` (what is on the system)

`#install_dir`
`#binary` (relies on `#install_dir`)

`#windows_version` (MSWebdriver only)

So as to the specifics referenced:
`Webdrivers.chrome.version` ==> we can do `Webdrivers::Chromedriver.chrome_version` if we make the private `#chrome_version` a public method

`Webdrivers.chrome.driver_version ` ==> this is the same as `Webdrivers::Chromedriver.current_version ` 

`Webdrivers.chrome.driver_path` ==> this is the same as `Webdrivers::Chromedriver.binary `; `#driver_path` probably makes more sense than `#binary`, so maybe we should rename it.

kapoorlakshya:
`version` and `current_version` seem synonymous and could cause confusion at first look. Maybe it'll be more intuitive if we replace `versions` with `desired_version` as the public method? Then we'll have `desired_version` (what the user wants) and `current_version` (what the user has).

titusfortner:
00ebdc2b6a3aec11b99d8fd3bfec0d98cb8d3137

PR #86, opened 2019-05-02
add specs with updated implementation

### LINKED ISSUE #80: Avoid HTTP check if version installed matches version pinned in config
My config
```ruby
require "webdrivers"
Webdrivers::Chromedriver.version = '74.0.3729.6'
Webdrivers.logger.level = :DEBUG
```

Output
```sh
➜ bin/rails test:system
Run options: --seed 1780

# Running:

Capybara starting Puma...
* Version 3.12.1 , codename: Llamas in Pajamas
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:51226
2019-04-29 12:58:13 DEBUG Webdrivers Looking for Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:58:14 DEBUG Webdrivers Get response: #<Net::HTTPOK 200 OK readbody=true>
2019-04-29 12:58:14 DEBUG Webdrivers Found Site: https://chromedriver.storage.googleapis.com
2019-04-29 12:58:14 DEBUG Webdrivers Checking current version
2019-04-29 12:58:14 DEBUG Webdrivers File is already downloaded: true
2019-04-29 12:58:14 DEBUG Webdrivers Current /Users/edgar/.webdrivers/chromedriver version: ChromeDriver 74.0.3729.6 (255758eccf3d244491b8a1317aa76e1ce10d57e9-refs/branch-heads/3729@{#29})

2019-04-29 12:58:14 DEBUG Webdrivers Expected webdriver version found
```

titusfortner:
This would fix #80 

Please make sure that these specs match what you expect. Chrome is a little special, but IE & Firefox are the exact same logic.

~~I updated specs for all of the drivers except for Edge. The existing specs are passing for Edge, but I don't think this code (or existing code works correctly for Edge). I'm going to propose some code changes for that separately before release.~~
I added specs with separate logic for Edge. Things are a little weird with the latest updates: https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/
This will need to be updated once we see the pattern that `msedgedriver.exe` takes going forward

1. Adds Custom Errors
2. Minimizes unnecessary network calls
3. Abstracts out a couple methods to make testing easier
4. Chromedriver code is slightly more strict when in offline mode (`#sufficient_binary?`)

Also, I want to release a 3.9 before merging this in.

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'nokogiri'
 require 'shellwords'
 require 'webdrivers/common'
 
@@ -11,19 +10,17 @@ module Webdrivers
         Webdrivers.logger.debug 'Checking current version'
         return nil unless exists?
 
-        ver = `#{binary} --version`
-        Webdrivers.logger.debug "Current #{binary} version: #{ver}"
+        version = binary_version
+        return nil if version.nil?
 
         # Matches 2.46, 2.46.628411 and 73.0.3683.75
-        normalize_version ver[/\d+\.\d+(\.\d+)?(\.\d+)?/]
+        normalize_version version[/\d+\.\d+(\.\d+)?(\.\d+)?/]
       end
 
       def latest_version
         @latest_version ||= begin
-          raise StandardError, 'Can not reach site' unless site_available?
-
           # Versions before 70 do not have a LATEST_RELEASE file
-          return normalize_version('2.46') if release_version < normalize_version('70.0.3538')
+          return normalize_version('2.41') if release_version < normalize_version('70')
 
           latest_applicable = latest_point_release(release_version)
 
@@ -45,7 +42,7 @@ module Webdrivers
           msg = version > latest_release ? 'you appear to be using a non-production version of Chrome; ' : ''
           msg = "#{msg}please set `Webdrivers::Chromedriver.version = <desired driver version>` to an known "\
 'chromedriver version: https://chromedriver.storage.googleapis.com/index.html'
-          raise StandardError, msg
+          raise VersionError, msg
         end
       end
 
@@ -97,7 +94,7 @@ module Webdrivers
                 raise NotImplementedError, 'Your OS is not supported by webdrivers gem.'
               end.chomp
 
-        raise StandardError, 'Failed to find Chrome binary or its version.' if ver.nil? || ver.empty?
+        raise VersionError, 'Failed to find Chrome binary or its version.' if ver.nil? || ver.empty?
 
         Webdrivers.logger.debug "Browser version: #{ver}"
         normalize_version ver[/\d+\.\d+\.\d+\.\d+/] # Google Chrome 73.0.3683.75 -> 73.0.3683.75
@@ -106,7 +103,7 @@ module Webdrivers
       def chrome_on_windows
         if browser_binary
           Webdrivers.logger.debug "Browser executable: '#{browser_binary}'"
-          return `powershell (Get-ItemProperty '#{browser_binary}').VersionInfo.ProductVersion`.strip
+          return system_call("powershell (Get-ItemProperty '#{browser_binary}').VersionInfo.ProductVersion").strip
         end
 
         # Workaround for Google Chrome when using Jruby on Windows.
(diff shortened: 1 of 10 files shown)


lib/webdrivers/chromedriver.rb
```diff
@@ -44,7 +41,7 @@ def latest_point_release(version)
           msg = version > latest_release ? 'you appear to be using a non-production version of Chrome; ' : ''
           msg = "#{msg}please set `Webdrivers::Chromedriver.version = <desired driver version>` to an known "\
```
twalpole, line 43: Needs to be update if swapping to `required_version`

lib/webdrivers/mswebdriver.rb
```diff
@@ -1,38 +1,84 @@
...
 
-      # Webdriver binaries for Microsoft Edge are not backwards compatible.
-      # For this reason, instead of downloading the latest binary we download the correct one for the
-      # currently installed browser version.
-      alias version windows_version
+      def remove
+        @download_url = nil
+        @latest_version = nil
+        Webdrivers.logger.debug 'Deleting MicrosoftWebDriver.exe'
+        FileUtils.rm_f(File.join(install_dir, 'MicrosoftWebDriver.exe'))
```
twalpole: The _f version doesn't return errors if file can't be deleted - Not sure if that's a concern or not

spec/webdrivers/chromedriver_spec.rb
```diff
@@ -5,125 +5,207 @@
...
-  it 'downloads latest release for current version of Chrome by default' do
-    chromedriver.remove
-    chromedriver.download
-    cur_ver    = chromedriver.current_version.version
-    latest_ver = chromedriver.latest_version.version
-    expect(cur_ver).to eq latest_ver
-  end
+      it 'does not download when offline, binary exists and is less than v70' do
+        allow(Net::HTTP).to receive(:get_response).and_raise(SocketError)
+        allow(chromedriver).to receive(:downloaded?).and_return(true)
```
twalpole: Not 100% sure but wouldn't this stop it from being downloaded - and not actually test that it wasn't downloaded

spec/webdrivers/chromedriver_spec.rb
```diff
@@ -5,125 +5,207 @@
...
-  end
+        expect(File.exist?(chromedriver.binary)).to be false
+      end
 
-  it 'removes chromedriver' do
-    chromedriver.remove
-    expect(chromedriver.current_version).to be_nil
+      it 'does not download when offline, binary exists and matches major browser version' do
+        allow(Net::HTTP).to receive(:get_response).and_raise(SocketError)
+        allow(chromedriver).to receive(:downloaded?).and_return(true)
```
twalpole: Same concern as previous - again not sure I'm correct on this though

spec/webdrivers/chromedriver_spec.rb
```diff
@@ -5,125 +5,207 @@
...
+        allow(chromedriver).to receive(:current_version).and_return(Gem::Version.new('73.0.3683.20'))
+
+        chromedriver.update
+
+        expect(File.exist?(chromedriver.binary)).to be false
+      end
+
+      it 'raises ConnectionError when offline, and binary does not match major browser version' do
+        allow(Net::HTTP).to receive(:get_response).and_raise(SocketError)
+        allow(chromedriver).to receive(:downloaded?).and_return(true)
```
twalpole: Same

spec/webdrivers/chromedriver_spec.rb
```diff
@@ -5,125 +5,207 @@
...
+      allow(chromedriver).to receive(:chrome_version).and_return '71.0.3578.9999'
+
+      expect(chromedriver.latest_version).to eq Gem::Version.new('71.0.3578.137')
     end
 
-    it 'raises an exception for beta version' do
+    it 'raises VersionError for beta version' do
       allow(chromedriver).to receive(:chrome_version).and_return('100.0.0')
       msg = 'you appear to be using a non-production version of Chrome; please set '\
 '`Webdrivers::Chromedriver.version = <desired driver version>` to an known chromedriver version: '\
```
twalpole, line 131: `required_version` if changing

twalpole (commented):
After a quick read through I made a couple of comments.  The biggest thing I see is the use of `allow(...).to receive(:downloaded?).and_return(true)` in the tests that are verifying downloads don't happen.  I'm not 100% correct it's wrong but the naming seems like it could be.  I'll have to read more through the code to see exactly what happens with `downloaded?`

spec/webdrivers/chromedriver_spec.rb
titusfortner: This is testing if it is offline, and a binary is there, it doesn't still download.
setting `#downloaded?` to true is mocking that something is there. Then if the file doesn't actually exist, then we know it wasn't still downloaded. It's just stepping through the conditional in the method to make sure it short-circuits.

lib/webdrivers/chromedriver.rb
titusfortner: I was going to address everything from #36 is a separate PR

lib/webdrivers/mswebdriver.rb
kapoorlakshya: Yeah, same as #87.

titusfortner: Removed this code

lib/webdrivers/common.rb
```diff
@@ -9,45 +9,21 @@ class << self
...
-        # Can't find desired and no existing binary
-        if desired_version.nil?
-          msg = "Unable to find the latest version of #{file_name}; try downloading manually " \
-"from #{base_url} and place in #{install_dir}"
-          raise StandardError, msg
-        end
-
         if correct_binary?
-          Webdrivers.logger.debug 'Expected webdriver version found'
+          Webdrivers.logger.debug 'The desired webdriver version is already on the system'
```
kapoorlakshya: Maybe replace "desired" with "required" to match `required_version`?

lib/webdrivers/mswebdriver.rb
```diff
@@ -1,32 +1,46 @@
 # frozen_string_literal: true
 
+require 'nokogiri'
+
 module Webdrivers
   class MSWebdriver < Common
     class << self
+      attr_accessor :ignore
```
kapoorlakshya: What's the purpose of this?

titusfortner: I'm not sure MS webdriver works with webdrivers. This is a toggle for turning off the gem functionality just for Edge.

kapoorlakshya: I was not sure of the use case, but I get it now after looking at #102.

kapoorlakshya (approved):
Let's merge once specs pass.

PR #87, opened 2019-05-03
#remove - Fix an intermittent file locking issue on Windows.

kapoorlakshya:
Fixes an intermittent file locking issue in specs when `#remove` fails to remove the binary on Windows (Appveyor). The spec expects the file to be removed, but it is not because `FileUtils.rm_f` fails to delete and does not raise any errors.

Example: https://ci.appveyor.com/project/kapoorlakshya/webdrivers/builds/24270606

The solution in this PR is to use `File#delete` and then rescue `Errno::EACCES` and retry up to 3 times.

@twalpole Do you have any suggestions to further improve this?

### DIFF
--- a/lib/webdrivers/common.rb
+++ b/lib/webdrivers/common.rb
@@ -51,9 +51,19 @@ module Webdrivers
       end
 
       def remove
+        max_attempts  = 3
+        attempts_made = 0
+        delay         = 0.5
         Webdrivers.logger.debug "Deleting #{binary}"
         @download_url = nil
-        FileUtils.rm_f binary
+
+        begin
+          attempts_made += 1
+          File.delete binary if File.exist? binary
+        rescue Errno::EACCES # Solves an intermittent file locking issue on Windows
+          sleep(delay)
+          retry if File.exist?(binary) && attempts_made <= max_attempts
+        end
       end
 
       def download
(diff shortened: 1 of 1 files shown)


twalpole:
I believe this is caused by Windows not actually freeing up closed files until garbage collection is run. So a different solution could be to run `GC.start` before attempting the file deletion.  That would also potentially slightly slow down non-windows runs so not really sure it's any better than retrying like this PR does.

twalpole (approved):
Seems fine to me

twalpole:
@kapoorlakshya Actually - just noticed one thing -- if it doesn't succeed after the number of retries - it will just hide the error -- Probably better to reraise after the retry

lib/webdrivers/common.rb
```diff
@@ -51,9 +51,19 @@ def latest_version
...
         @download_url = nil
-        FileUtils.rm_f binary
+
+        begin
+          attempts_made += 1
+          File.delete binary if File.exist? binary
+        rescue Errno::EACCES # Solves an intermittent file locking issue on Windows
+          sleep(delay)
+          retry if File.exist?(binary) && attempts_made <= max_attempts
+        end
```
titusfortner, line 67: line 66 needs to have `raise` so this doesn't just ignore the error after max retries.

titusfortner (approved):
Go ahead and merge this & #88  after the one fix.

PR #88, opened 2019-05-03
Run specs on Appveyor.

### LINKED ISSUE #85: Implement Appveyor support
Need to be able to run the specs on Windows as well so we don't break things.

kapoorlakshya:
Closes #85. Will rebase this once #87 is merged.

https://ci.appveyor.com/project/kapoorlakshya/webdrivers/builds/24273630


titusfortner (approved):
This is good, pending we agree on the right solution from #87

PR #90, opened 2019-05-04
deprecate MSWebdriver

titusfortner:
I'm not certain the current implementation of MSWebdriver works for Edge versions < 17, and I can't get any of my virtual machines to even run the Microsoft Webdriver, let alone validate that the webdrivers gem logic is working. So, I'd like to just deprecate this entire class for webdrivers 4, and then create a new one to support the new Chromium version for 4.0. Note that Selenium 4 will not work with Edge 17 & below.

Any objections to this approach?

### DIFF
--- a/lib/webdrivers/mswebdriver.rb
+++ b/lib/webdrivers/mswebdriver.rb
@@ -3,6 +3,14 @@
 module Webdrivers
   class MSWebdriver < Common
     class << self
+      def update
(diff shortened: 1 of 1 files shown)


kapoorlakshya (approved):
Let's do it!

titusfortner:
I think it is going to make more sense to just delete this class entirely for 3.9, so I'm going to close this.

Issue #67, opened 2019-04-15
NoMethodError in Chrome (undefined method `strip' for nil:NilClass)


clockworkpc:
### System
Ubuntu 18.04
Ruby 2.5.1p57 (freshly installed)
NOTE: I also tried this on Ruby 2.6.2 and 2.5.5
Chromium version: Version 73.0.3683.75 (Official Build) Built on Ubuntu , running on Ubuntu 18.04 (64-bit)

### Steps Followed
Created a brand new rails app with default settings.
Here is my Gemfile:
[Gemfile.txt](https://github.com/titusfortner/webdrivers/files/3078943/Gemfile.txt)

### Chrome Browser (Fails)
```
Running via Spring preloader in process 29442
Loading development environment (Rails 5.2.3)
irb(main):001:0> Watir::Browser.new(:chrome)
rails_console: No such file or directory - --product-version
Traceback (most recent call last):
        2: from (irb):1
        1: from (irb):1:in `new'
NoMethodError (undefined method `strip' for nil:NilClass)
```

### Firefox Browser (Works OK)
```
irb(main):003:0> b = Watir::Browser.new(:firefox)
=> #<Watir::Browser:0x3a74db97bfeb780e url="about:blank" title="">
irb(main):005:0> b.goto('www.google.com')
=> "http://www.google.com"
irb(main):006:0>
```

### Current Solution: Go back to `chromedriver-helper` for Chrome
If I include `gem 'chromedriver-helper'` in my Gemfile, run `bundle install` and jump back into the Rails console, then Chrome works too.

```
Running via Spring preloader in process 29466
Loading development environment (Rails 5.2.3)
irb(main):001:0> b = Watir::Browser.new(:chrome)
=> #<Watir::Browser:0x..f9dd5c620c38cd000 url="data:," title="">
irb(main):002:0> b.goto('www.google.com')
=> "http://www.google.com"
```

Firefox is fine either way.

kapoorlakshya:
Hi @clockworkpc, please try providing the path to the Chromium binary as described [here](https://github.com/titusfortner/webdrivers#when-using-chromechromium).

clockworkpc:
> Hi @clockworkpc, please try providing the path to the Chromium binary as described [here](https://github.com/titusfortner/webdrivers#when-using-chromechromium).

Thanks for the reply.  I checked my Chromium path and found that it indeed differs from the standard installation path:

```
$ which chromium-browser 
/usr/bin/chromium-browser
```

[...]

PR #91, opened 2019-05-05
change public names to more intelligent defaults

titusfortner:
This addresses #36 and is based on #86

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -29,6 +29,16 @@ module Webdrivers
         end
       end
 
+      # Returns currently installed Chrome version
+      def chrome_version
+        ver = send("chrome_on_#{platform}").chomp
+
+        raise VersionError, 'Failed to find Chrome binary or its version.' if ver.nil? || ver.empty?
+
+        Webdrivers.logger.debug "Browser version: #{ver}"
+        normalize_version ver[/\d+\.\d+\.\d+\.\d+/] # Google Chrome 73.0.3683.75 -> 73.0.3683.75
+      end
+
       private
 
       def latest_point_release(version)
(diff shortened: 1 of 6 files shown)


lib/webdrivers/chromedriver.rb
```diff
@@ -115,14 +114,14 @@ def chrome_on_windows
         end
 
         # Default to Google Chrome
-        reg        = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\chrome.exe'
+        reg = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths\chrome.exe'
         executable = system_call("powershell (Get-ItemProperty '#{reg}' -Name '(default)').'(default)'").strip
         Webdrivers.logger.debug "Browser executable: '#{executable}'"
         ps = "(Get-Item (Get-ItemProperty '#{reg}').'(default)').VersionInfo.ProductVersion"
         system_call("powershell #{ps}").strip
```
kapoorlakshya, line 122: I think we can remove `strip` from these methods to avoid the error message reported in #67. And it would be better to perform the nil check and raise `VersionError` before we `chomp` on line 33.

kapoorlakshya: Same on line 133 and 139.

titusfortner: Since that line isn't part of this PR, let's resolve this for now and do a sweep for any unnecessary commands before we release 3.9.

kapoorlakshya: Sounds good!

lib/webdrivers/chromedriver.rb
```diff
@@ -28,6 +28,16 @@ def latest_version
         @latest_version = normalize_version(latest_applicable)
       end
 
+      # Returns currently installed Chrome version
+      def chrome_version
+        ver = send("chrome_on#{platform}").chomp
```
kapoorlakshya: Looks like a `_` is missing. Should be `chrome_on_#{platform}`.

kapoorlakshya (approved):
Change looks good overall. Let's merge once specs pass.

titusfortner:
merged: 00ebdc2b6a3aec11b99d8fd3bfec0d98cb8d3137

Issue #77, opened 2019-04-28
parallelization issues


beauraF:
Hello,

_Wedrivers_ is having some problems with parallelization. This makes it unusable on _Heroku CI_ and under _Rails 6_.
One of the threads downloads the driver while the others try to access the resource. Which leads to this error.

```
Error:
AuthenticationsTest#test_successful_login:
Zip::Error: File /home/circleci/.webdrivers/chromedriver_linux64.zip has zero size. Did you mean to pass the create flag?
    test/system/authentications_test.rb:7:in `block in <class:AuthenticationsTest>'


rails test test/system/authentications_test.rb:6
```

```
Error:
AuthenticationsTest#test_failed_login:
ChildProcess::LaunchError: Text file busy - /home/circleci/.webdrivers/chromedriver
    test/system/authentications_test.rb:14:in `block in <class:AuthenticationsTest>'


rails test test/system/authentications_test.rb:13

Traceback (most recent call last):
	13: from /home/circleci/doctopod/vendor/bundle/gems/minitest-5.11.3/lib/minitest.rb:63:in `block in autorun'
	12: from /home/circleci/doctopod/vendor/bundle/gems/minitest-5.11.3/lib/minitest.rb:133:in `run'
	11: from /home/circleci/doctopod/vendor/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/testing/parallelization.rb:70:in `start'
	10: from /home/circleci/doctopod/vendor/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/testing/parallelization.rb:70:in `map'
	 9: from /home/circleci/doctopod/vendor/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/testing/parallelization.rb:70:in `each'
	 8: from /home/circleci/doctopod/vendor/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/testing/parallelization.rb:70:in `times'
	 7: from /home/circleci/doctopod/vendor/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/testing/parallelization.rb:71:in `block in start'
	 6: from /home/circleci/doctopod/vendor/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/testing/parallelization.rb:71:in `fork'
	 5: from /home/circleci/doctopod/vendor/bundle/gems/selenium-webdriver-3.142.0/lib/selenium/webdriver/common/platform.rb:150:in `block in exit_hook'
[...]

Issue #30, opened 2019-03-03
Use Static URL for geckodriver


titusfortner:
We can parse with this https://github.com/mozilla/geckodriver/releases/latest

titusfortner:
eb7f3d4a14c84ba957a5fcbd46388479f080a4e0

PR #94, opened 2019-05-05
3.9

titusfortner:
Ok, this is everything I think we need as a final 3.x release.
This might be a pain to review, but I tried to make each commit obvious and straightforward.

This is all dependent on the refactoring in #86, #90 & #91 
Caching addresses #29 & #77 
Direct downloading of geckodriver addresses #30 

I've got cache time defaulting to zero, with the idea of making it default for 4.x. Or should we set a value for 3.9?

I'm thinking we release this and then for 4.0 we:
1. Add the documentation
2. Create the Rake Tasks
3. Remove the deprecations
4. Change default cache_time to 24 hours

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,8 @@
 require: rubocop-rspec
 
+Layout/SpaceInsideHashLiteralBraces:
+  EnforcedStyle: no_space
+
 Layout/EndOfLine:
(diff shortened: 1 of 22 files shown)


titusfortner:
Only thing left to evaluate is #104 so closing this.

PR #97, opened 2019-05-06
Rename `downloaded?` to `exists?` to better identify what it does

twalpole:


titusfortner:
private method name change makes sense.

PR #102, opened 2019-05-06
remove support for Microsoft WebDriver for Edge browser

titusfortner:
I think this is the correct approach for warning existing users.

Throw a warning with instructions on how to fix and how to ignore the warning.

Note that Selenium 3.142.0 broke looking at `#driver_path` so I need to get that figured out as well.

### DIFF
--- a/lib/webdrivers/mswebdriver.rb
+++ b/lib/webdrivers/mswebdriver.rb
@@ -3,72 +3,30 @@
 require 'webdrivers/common'
 
 module Webdrivers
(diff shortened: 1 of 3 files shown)


kapoorlakshya:
Not sure how many people are using Edge, but this could be unexpected in a non-major release. The existing code still downloads the driver (tested on Edge 16), so I would prefer to only deprecate it. 

Can this wait till v4?

kapoorlakshya (approved):

PR #104, opened 2019-05-07
implements driver caching

titusfortner:
This creates a `#{driver_path}.version` file, and we read its modified date on the first lookup. There might still be a race condition, but I think it'll solve #77 well enough?

The idea is to set the default value in 3.x to 0, but allow people to set it if they want.
Then 4.x will be set by default to 1 day (86_400)

Note that for Chrome if a major browser version update has happened, it will ignore the cache, so effectively this is only preventing using a newer point release up to the expiration of the `cache_time`

### DIFF
--- a/lib/webdrivers/common.rb
+++ b/lib/webdrivers/common.rb
@@ -17,6 +17,12 @@ module Webdrivers
   class << self
     attr_accessor :proxy_addr, :proxy_port, :proxy_user, :proxy_pass, :install_dir
 
+    attr_writer :cache_time
+
+    def cache_time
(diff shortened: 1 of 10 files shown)


lib/webdrivers/common.rb
```diff
@@ -120,7 +124,7 @@ def sufficient_binary?
       end
 
       def normalize_version(version)
-        Gem::Version.new(version.to_s)
+        Gem::Version.new(version.nil? ? nil : version.to_s)
```
twalpole: version&.to_s

PR #105, opened 2019-05-08
locate Chrome the same way as chromedriver 

### LINKED ISSUE #45: Locate Chrome the same way as Google
Here is Google's code for how they get the default browser.
https://cs.chromium.org/chromium/src/chrome/test/chromedriver/chrome/chrome_finder.cc
This won't be an issue in 99% of cases, but it is possible that the webdrivers method and the chromedriver method can each get different information resulting in the wrong driver being used.

titusfortner:
Fixes #45 

I think I got this working the same was as Google.
It appears to work on Linux, Windows, JRuby & Mac for the testing environments we have at least.

Would it make sense to fall back to what we were doing before? chromedriver does not do anything with ENV or Registry like we were doing, though.

Since it's a potentially breaking change, the idea is to release this in a 4.0.alpha and get feedback from people as to whether it's working for them.

### DIFF
--- a/lib/webdrivers/chrome_finder.rb
+++ b/lib/webdrivers/chrome_finder.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module Webdrivers
+  class ChromeFinder
+    class << self
+      def version
+        location = Selenium::WebDriver::Chrome.path || send("#{System.platform}_location")
+        version = send("#{System.platform}_version", location)
+
+        raise VersionError, 'Failed to find Chrome binary or its version.' if version.nil? || version.empty?
+
+        Webdrivers.logger.debug "Browser version: #{version}"
+        version[/\d+\.\d+\.\d+\.\d+/] # Google Chrome 73.0.3683.75 -> 73.0.3683.75
+      end
+
+      def win_location
+        return Selenium::WebDriver::Chrome.path unless Selenium::WebDriver::Chrome.path.nil?
+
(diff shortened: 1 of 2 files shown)


lib/webdrivers/chrome_finder.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+        end
+      end
+
+      def win_version(location)
+        System.call "powershell (Get-ItemProperty '#{location}').VersionInfo.ProductVersion"
+      end
+
+      def linux_version(location)
+        System.call("#{Shellwords.escape location} --product-version").strip
```
kapoorlakshya: `&.strip` could be helpful in preventing `NoMethodError (undefined method 'strip' for nil:NilClass)` if Chrome is installed in a custom location.

lib/webdrivers/chrome_finder.rb
```diff
@@ -0,0 +1,63 @@
...
+        directories.each do |dir|
+          files.each do |file|
+            option = "#{dir}/#{file}"
+            return option if File.exist?(option)
+          end
+        end
+      end
+
+      def win_version(location)
+        System.call "powershell (Get-ItemProperty '#{location}').VersionInfo.ProductVersion"
```
kapoorlakshya: Missing `strip`.

kapoorlakshya (approved):
I'm glad this is finally in place.

titusfortner:
I'm going to hold off merging this until after a 3.9.x bug fixes

Issue #28, opened 2019-01-02
Rake Tasks for Update & Remove


titusfortner:
Webdrivers gem should support a means for deleting binary without requiring the user to `rm -rf ~/.webdrivers`

twalpole:
If providing this I think it should a rake task rather than a public API code method

titusfortner:
I implemented `#remove` a long time ago, so I'm not sure what this was about, but we can track the Rake tasks with it now.

PR #106, opened 2019-05-08
ensure the driver file is also present after validating the cache file

titusfortner:
This is kind of whack-a-mole with the race conditions for #77...

User is reporting `ChildProcess::LaunchError: Text file busy`, which means that it is attempting to start a file that is currently open for writing. 

The scenario I'm picturing for 2 processes running in parallel:
1 - Sees no version file and creates it
2 - Sees a version file and continues on
1 - Downloads / Extracts driver
2 - Tries to use the driver before it is finished being downloaded/extracted

Waiting for the driver to exist after validating that it should not be downloaded again should be sufficient for this particular race condition.

The tricky part here is determining how long it should take for a driver to be downloaded and extracted, etc, and how to recover from an unforeseen state.

I don't think there's going to be a good way to completely solve this 100%. The right answer might be the rake tasks in #28

### DIFF
--- a/lib/webdrivers/system.rb
+++ b/lib/webdrivers/system.rb
@@ -39,10 +39,20 @@ module Webdrivers
       end
 
       def valid_cache?(file_name)
-        file = "#{install_dir}/#{file_name.gsub('.exe', '')}.version"
-        return false unless File.exist?(file)
+        driver_file = "#{install_dir}/#{file_name.gsub('.exe', '')}"
+        cache_file = "#{driver_file}.version"
+        return false unless File.exist?(cache_file)
 
-        Time.now - File.mtime(file) < Webdrivers.cache_time
+        return false if Time.now - File.mtime(cache_file) < Webdrivers.cache_time
+
+        begin
+          Selenium::WebDriver::Wait.new(timeout: 30, interval: 0.1).until do
+            File.exist?(driver_file)
+          end
+          return true
+        rescue TimeoutError
+          return false
+        end
       end
 
       def download(url, target)
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -24,7 +24,7 @@ Metrics/BlockLength:
     - 'spec/**/*'
 
 Metrics/ClassLength:
-  Max: 116
+  Max: 119
 
 Metrics/CyclomaticComplexity:
   Max: 8


twalpole:
I guess the full solution to this would be to go to a cross process mutex solution. Easy enough in non-windows using `flock` - I’m not sure how to implement on windows though

titusfortner:
We could have a windows and non-windows solutions. I suspect that most people wanting to run parallel aren't doing it on windows.

Also, I think I'm going to have to refactor a bunch of things in the System class to make the mocks I'm using more reasonable, to get the specs to pass with this PR.

twalpole:
Isn't the better short term solution for parallel to just set a different install dir for each process? That removes all the race conditions. Yes it would download the driver multiple times, but that time should be negligible in cases where parallel testing is providing the most benefit.

lib/webdrivers/system.rb
```diff
@@ -39,10 +39,20 @@ def cached_version(file_name)
...
-        return false unless File.exist?(file)
+        driver_file = "#{install_dir}/#{file_name.gsub('.exe', '')}"
+        cache_file = "#{driver_file}.version"
+        return false unless File.exist?(cache_file)
 
-        Time.now - File.mtime(file) < Webdrivers.cache_time
+        return false if Time.now - File.mtime(cache_file) < Webdrivers.cache_time
+
+        begin
+          Selenium::WebDriver::Wait.new(timeout: 30, interval: 0.1).until do
```
twalpole, line 49: If we're going to do this should probably make the wait time configurable

titusfortner: which is part of why the more I think about it, the less I like this approach.

titusfortner:
A different install directory per process would mean that you would end up downloading a new driver every single execution, and we'd potentially have trouble cleaning up afterwards. I'd almost rather punt on this whole thing and require people to use a rake task ahead of the parallel execution.

kapoorlakshya:
> I'd almost rather punt on this whole thing and require people to use a rake task ahead of the parallel execution.

I think that might be a better or at least an easier solution than trying to add OS specific code.

titusfortner:
the more I think about this the less I want to have this in the project, so I'm closing.

PR #110, opened 2019-05-09
Bump the selenium-webdriver version since older version 3.0.x did not…

tchia04:
… work

It was causing invalid argument on driver_path until I hard coded the version of selenium-webdriver to the latest


titusfortner:
Thanks for the submission!

This commit should address the underlying problem: fc447d790f868bbd4d3525cbf9c4c9d10b5b486d

I'll do a point release shortly, let me know if that doesn't fix it.

PR #112, opened 2019-05-09
Add method and class documentation.

### LINKED ISSUE #92: Better code documentation
All public methods need to get yardoc attention.

kapoorlakshya:
Closes #92.


titusfortner (changes requested):
Let's add in return values for these methods as well:
`@return [Gem::Version]` or File or String or URL or whatever it is.

PR #117, opened 2019-05-14
Add Rake tasks for update, remove, and version.

### LINKED ISSUE #28: Rake Tasks for Update & Remove

kapoorlakshya:
```
rake webdrivers:chromedriver:remove  # Removes chromedriver
rake webdrivers:chromedriver:update  # Updates chromedriver
rake webdrivers:geckodriver:remove   # Removes geckodriver
rake webdrivers:geckodriver:update   # Updates geckodriver
rake webdrivers:iedriver:remove      # Removes IEServerDriver
rake webdrivers:iedriver:update      # Updates IEServerDriver
```

Closes #28.

### DIFF
--- a/lib/webdrivers/system.rb
+++ b/lib/webdrivers/system.rb
@@ -26,7 +26,7 @@ module Webdrivers
       end
 
       def install_dir
-        Webdrivers.install_dir || File.expand_path(File.join(ENV['HOME'], '.webdrivers'))
+        Webdrivers.install_dir
       end
 
       def cache_version(file_name, version)
--- a/lib/webdrivers/tasks/chromedriver.rake
+++ b/lib/webdrivers/tasks/chromedriver.rake
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+namespace :webdrivers do
+  require 'webdrivers/chromedriver'
+
+  namespace :chromedriver do
+    Webdrivers.logger.level = :info
+
+    desc 'Print current chromedriver version'
+    task :version do
+      gem_ver = Webdrivers::Chromedriver.current_version
+      if gem_ver
+        Webdrivers.logger.info "chromedriver #{gem_ver.version}"
+      else
+        Webdrivers.logger.warn 'No existing chromedriver found.'
+      end
+    end
+
+    desc 'Remove and download updated chromedriver if necessary'
+    task :update, [:version] do |_, args|
+      args.with_defaults(version: 0)
+      Webdrivers.cache_time                     = ENV.fetch('WD_CACHE_TIME', 86_400)
+      Webdrivers.install_dir                    = ENV.fetch('WD_INSTALL_DIR', nil)
+      Webdrivers::Chromedriver.required_version = args.version
+      Webdrivers::Chromedriver.update
+      Webdrivers.logger.info "Updated to chromedriver #{Webdrivers::Chromedriver.current_version}"
+    end
+
+    desc 'Force remove chromedriver'
+    task :remove do
+      unless File.exist? Webdrivers::Chromedriver.driver_path
+        Webdrivers.logger.info 'No existing chromedriver to remove.'
+        next # Return early
+      end
+
+      cur_version = Webdrivers::Chromedriver.current_version
+      Webdrivers::Chromedriver.remove
+
+      if File.exist? Webdrivers::Chromedriver.driver_path # Failed for some reason
+        Webdrivers.logger.error 'Failed to remove chromedriver. Please try removing manually.'
+      else
+        Webdrivers.logger.info "Removed chromedriver #{cur_version}."
+      end
+    end
+  end
+end
--- a/lib/webdrivers/tasks/geckodriver.rake
+++ b/lib/webdrivers/tasks/geckodriver.rake
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+namespace :webdrivers do
+  require 'webdrivers/geckodriver'
+
+  namespace :geckodriver do
+    Webdrivers.logger.level = :info
+
+    desc 'Print current geckodriver version'
+    task :version do
+      gem_ver = Webdrivers::Geckodriver.current_version
+      if gem_ver
+        Webdrivers.logger.info "geckodriver #{gem_ver.version}"
+      else
+        Webdrivers.logger.warn 'No existing geckodriver found.'
+      end
+    end
+
+    desc 'Remove and download updated geckodriver if necessary'
+    task :update, [:version] do |_, args|
+      args.with_defaults(version: 0)
+      Webdrivers.cache_time                    = ENV.fetch('WD_CACHE_TIME', 86_400)
+      Webdrivers.install_dir                   = ENV.fetch('WD_INSTALL_DIR', nil)
+      Webdrivers::Geckodriver.required_version = args.version
+      Webdrivers::Geckodriver.update
+      Webdrivers.logger.info "Updated to geckodriver #{Webdrivers::Geckodriver.current_version}"
+    end
+
+    desc 'Force remove geckodriver'
+    task :remove do
+      unless File.exist? Webdrivers::Geckodriver.driver_path
+        Webdrivers.logger.info 'No existing geckodriver to remove.'
+        next # Return early
+      end
+
+      cur_version = Webdrivers::Geckodriver.current_version
+      Webdrivers::Geckodriver.remove
+
+      if File.exist? Webdrivers::Geckodriver.driver_path # Failed for some reason
+        Webdrivers.logger.error 'Failed to remove geckodriver. Please try removing manually.'
+      else
+        Webdrivers.logger.info "Removed geckodriver #{cur_version}."
+      end
+    end
+  end
+end
--- a/lib/webdrivers/tasks/iedriver.rake
+++ b/lib/webdrivers/tasks/iedriver.rake
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+namespace :webdrivers do
+  require 'webdrivers/iedriver'
+
+  namespace :iedriver do
+    Webdrivers.logger.level = :info
+
+    desc 'Print current IEDriverServer version'
+    task :version do
+      gem_ver = Webdrivers::IEdriver.current_version
+      if gem_ver
+        Webdrivers.logger.info "IEDriverServer #{gem_ver.version}"
+      else
+        Webdrivers.logger.warn 'No existing IEDriverServer found.'
+      end
+    end
+
+    desc 'Remove and download updated IEDriverServer if necessary'
+    task :update, [:version] do |_, args|
+      args.with_defaults(version: 0)
+      Webdrivers.cache_time                 = ENV.fetch('WD_CACHE_TIME', 86_400)
+      Webdrivers.install_dir                = ENV.fetch('WD_INSTALL_DIR', nil)
+      Webdrivers::IEdriver.required_version = args.version
+      Webdrivers::IEdriver.update
+      Webdrivers.logger.info "Updated to IEDriverServer #{Webdrivers::IEdriver.current_version}"
+    end
+
+    desc 'Force remove IEDriverServer'
+    task :remove do
+      unless File.exist? Webdrivers::IEdriver.driver_path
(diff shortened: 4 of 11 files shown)


twalpole:
I think the update tasks  probably need to accept version numbers since any code specifying a version number to lock to won’t be loaded when these are run

twalpole:
Additionally the new tasks should not be in the main Rakefile - they should be in a separate file other projects require - See http://mdzhang.com/blog/code/2016/09/10/create-ruby-gem-that-adds-rake-tasks/ - ignore the part about separate files for each task though - that seems unnecessary

titusfortner:
We also want to be able to use caching with update.

I'm thinking we probably want to leverage `--cache-time` and `--required-version` parameters with an option parser (http://www.mikeball.info/blog/rake-option-parser/)

The description should be more clear; something like:
"force removal of driver"
and
"remove and download updated driver if necessary"

Info about this should also get added to the README

Finally, I think we want to do `Webdrivers.logger.info = 'message'` instead of `puts`

kapoorlakshya:
@twalpole @titusfortner I wasn't sure of the expectation at first, but your comments and links made them very clear. Thanks for that :)

Updated task descriptions:

```
$ bundle exec rake -T
rake webdrivers:chromedriver:remove  # Force remove chromedriver
rake webdrivers:chromedriver:update  # Remove and download updated chromedriver if necessary
rake webdrivers:geckodriver:remove   # Force remove geckodriver
rake webdrivers:geckodriver:update   # Remove and download updated geckodriver if necessary
rake webdrivers:iedriver:remove      # Force remove IEDriverServer
rake webdrivers:iedriver:update      # Remove and download updated IEDriverServer if necessary
```

Added options for `update` task:

```
-c, --cache-time SECONDS
-v, --required-version VERSION
```

`-t` is taken by `rake` for `--trace`. Feel free to suggest a better one for `--cache-time`.

`-h, --help` option:

```
$ bundle exec rake webdrivers:chromedriver:update -- -h
Usage: webdrivers::chromedriver::update [options]
    -c, --cache-time SECONDS         (Optional) Number of Seconds to wait between update checks
    -v, --required-version VERSION   (Optional) chromedriver version to download
```

`remove` task:

```
$ bundle exec rake webdrivers:chromedriver:remove
2019-05-14 23:10:29 INFO Webdrivers Removed chromedriver 74.0.3729.6.

$ bundle exec rake webdrivers:chromedriver:remove
2019-05-14 23:10:58 INFO Webdrivers No existing chromedriver to remove.
```
I would like to DRY up the `OptionParser` code, but the only way I can think of is wrapping it inside a   module (helper method) and passing the driver class name to it for the option descriptions. Do you have any strategies on how to make it cleaner, if needed?

I will update the `README` once we finalize on the code.

twalpole:
Rake can take multiple tasks on the same command line (`rake webdrivers:chromedriver:update spec` for instance), so switches don't generally work well.  The normal rake way to do it would be either using the built in parameter parsing - https://github.com/ruby/rake/blob/master/doc/rakefile.rdoc#tasks-with-arguments - which would then be called like `rake webdriver:chromedriver:update[73.0329.123.11]` or via environment variables `rake webdriver:chromedriver:update version="73.x.x.x"`(ENV['version'] would access the version).  Obviously the second setup has the same issue as using the switches like it's currently implemented ( You can't do `rake webdriver:chromedriver:update webdriver:geckodriver:update` because there would be no way to pass 2 versions)

twalpole:
Additionally please add Railtie support as described in the previous link I sent, so that for users of Rails the rake tasks get automatically loaded.

titusfortner:
@twalpole I haven't tested it, but wouldn't this work as expected?
```
bundle exec rake webdrivers:chromedriver:update spec -- --cache-time 30 --version 74.0.3729.6
```

Of note, users will also need to set `:cache_time` in the code as well so that it doesn't go back online during each individual test run following the rake task. So it might make sense to have this be an environment variable set in the rake task and have our code respect it.

twalpole:
@titusfortner that might, but what if my tests run with chromedriver and geckodriver - then you can't set multiple versions, however

    bundle exec rake webdrivers:chromedriver:update[74.0.3729.6] webdrivers:geckodriver:update[0.24.0] spec CACHE_TIME=30

would allow it (and simplifies the tasks by removing the need for OptionParser in favor of using rakes native features).

twalpole:
hmm - of course that then breaks in zsh by default and users need to escape the [ and ] on the command line - 🤷‍♂

kapoorlakshya:
Updated the tasks as per the discussion here and on Slack:

```bash
$ bundle exec rake -T
rake webdrivers:chromedriver:remove           # Force remove chromedriver
rake webdrivers:chromedriver:update[version]  # Remove and download updated chromedriver if necessary
rake webdrivers:geckodriver:remove            # Force remove geckodriver
rake webdrivers:geckodriver:update[version]   # Remove and download updated geckodriver if necessary
rake webdrivers:iedriver:remove               # Force remove IEDriverServer
rake webdrivers:iedriver:update[version]      # Remove and download updated IEDriverServer if necessary
```

The following works now and `CACHE_TIME` defaults to `86_400`:

```bash
$ bundle exec rake webdrivers:chromedriver:update[2.46] webdrivers:geckodriver:update[0.24.0] CACHE_TIME=30
2019-05-15 22:36:41 INFO Webdrivers Updated to chromedriver 2.46.628388
2019-05-15 22:36:41 INFO Webdrivers Updated to geckodriver 0.24.0
```

@twalpole Added `Railtie` support. Can you please test it when you get a chance?

titusfortner:
Yeah, this looks really good. I still like the OptionParser better as it is more explicit, and then just tell people to create their own rake file if they want to combine them, but whatever, let's do it the default Rake way. :)

Only two changes:
1. We should support `ENV['CACHE_TIME']` in the code as well as the Rake tasks.
2. I know it works, but we should also have a spec for things working when we set `required_version` equal to zero since that's what this Rake task is doing

twalpole:
Tested out the railties - seems to work fine

```
RAILS_ENV=test rake -T
Running via Spring preloader in process 35379
rake about                                    # List versions of all Rails frameworks and the environment
rake active_storage:install                   # Copy over the migration needed to the application
rake app:template                             # Applies the template supplied by LOCATION=(/path/to/template) or URL
...
rake tmp:create                               # Creates tmp directories for cache, sockets, and pids
rake webdrivers:chromedriver:remove           # Force remove chromedriver
rake webdrivers:chromedriver:update[version]  # Remove and download updated chromedriver if necessary
rake webdrivers:geckodriver:remove            # Force remove geckodriver
rake webdrivers:geckodriver:update[version]   # Remove and download updated geckodriver if necessary
rake webdrivers:iedriver:remove               # Force remove IEDriverServer
rake webdrivers:iedriver:update[version]      # Remove and download updated IEDriverServer if necessary
rake yarn:install                             # Install all JavaScript dependencies as specified via Yarn
```

One thing I did think about is the install dir - I think we probably want a way to specify that for these tasks too - thoughts?

kapoorlakshya:
@titusfortner I'll add support for `ENV['CACHE_TIME']` in the core code. And I am okay with adding a spec for `required_version = 0` or we can set it to `Gem::Version('')` as PR #108 added logic to handle it.

@twalpole Sounds reasonable. How about `bundle exec rake webdrivers:chromedriver:update[2.46] webdrivers:geckodriver:update[0.24.0] CACHE_TIME=30 INSTALL_DIR='my_dir'`? 

Additionally, I would like to rename `CACHE_TIME` to `WD_CACHE_TIME` (and `WD_INSTALL_DIR`) so it's a bit more specific to this gem via the "WD" notation. That way we can set it at the global level and avoid possible conflicts with any other `INSTALL_DIR` or caching related ENV variables from other tools/user's project.

kapoorlakshya:
Note to myself:

- [x] Implement support for `ENV['WD_CACHE_TIME']` and `ENV['WD_INSTALL_DIR']` in the core code.
- [x] Update rake tasks to use `WD_CACHE_TIME` 
- [x] Add specs to test `required_version` as `nil` and `0`. Latest version should be downloaded.
- [x] Implement support for `WD_INSTALL_DIR` in update rake task
- [x] Add rake task `webdrivers:chromedriver:version` to print `current_version`
- [x] Update `README`.

kapoorlakshya:
Here's the updated list:

```bash
$ bundle exec rake -T
rake webdrivers:chromedriver:remove           # Force remove chromedriver
rake webdrivers:chromedriver:update[version]  # Remove and download updated chromedriver if necessary
rake webdrivers:chromedriver:version          # Print current chromedriver version
rake webdrivers:geckodriver:remove            # Force remove geckodriver
rake webdrivers:geckodriver:update[version]   # Remove and download updated geckodriver if necessary
rake webdrivers:geckodriver:version           # Print current geckodriver version
rake webdrivers:iedriver:remove               # Force remove IEDriverServer
rake webdrivers:iedriver:update[version]      # Remove and download updated IEDriverServer if necessary
rake webdrivers:iedriver:version              # Print current IEDriverServer version
```
`:version` task:

```bash
$ bundle exec rake webdrivers:chromedriver:version
2019-05-18 17:45:59 INFO Webdrivers chromedriver 74.0.3729.6
```

lib/webdrivers/system.rb
```diff
@@ -26,7 +26,7 @@ def delete(file)
       end
 
       def install_dir
-        Webdrivers.install_dir || File.expand_path(File.join(ENV['HOME'], '.webdrivers'))
+        Webdrivers.install_dir
```
titusfortner, line 29: is this different from just using the `attr_accessor`?

lib/webdrivers/tasks/chromedriver.rake
```diff
@@ -0,0 +1,46 @@
...
+  require 'webdrivers/chromedriver'
+
+  namespace :chromedriver do
+    Webdrivers.logger.level = :info
+
+    desc 'Print current chromedriver version'
+    task :version do
+      gem_ver = Webdrivers::Chromedriver.current_version
+      if gem_ver
+        Webdrivers.logger.info 'chromedriver ' + gem_ver.version
```
titusfortner: Use `"#{}"` syntax for concatenation

lib/webdrivers/tasks/geckodriver.rake
```diff
@@ -0,0 +1,46 @@
...
+  require 'webdrivers/geckodriver'
+
+  namespace :geckodriver do
+    Webdrivers.logger.level = :info
+
+    desc 'Print current geckodriver version'
+    task :version do
+      gem_ver = Webdrivers::Geckodriver.current_version
+      if gem_ver
+        Webdrivers.logger.info 'geckodriver ' + gem_ver.version
```
titusfortner: Use "#{}" syntax for concatenation

lib/webdrivers/tasks/iedriver.rake
```diff
@@ -0,0 +1,46 @@
...
+  require 'webdrivers/iedriver'
+
+  namespace :iedriver do
+    Webdrivers.logger.level = :info
+
+    desc 'Print current IEDriverServer version'
+    task :version do
+      gem_ver = Webdrivers::IEdriver.current_version
+      if gem_ver
+        Webdrivers.logger.info 'IEDriverServer ' + gem_ver.version
```
titusfortner: Use "#{}" syntax for concatenation

titusfortner (approved):
LGTM. Couple notes, then I'm cool with merging. Next release will be a beta to make sure it works for users, so they can let us know if there are issues or use cases we overlooked... :)

titusfortner:
Oh, this also needs a README update to discuss ENV specifics & Available tasks

lib/webdrivers/system.rb
kapoorlakshya: The accessor (reader method) was only returning the user define path or `nil`. I attempted to move all the logic in `System.install_dir` and have `Webdrivers.install_dir` reference it, but then that created a circular reference as the first check in System is  for `Webdrivers.install_dir`, which itself was referencing `System.install_dir` to get the default path.

kapoorlakshya:
Made requested changes and updated `README`.

PR #120, opened 2019-05-14
Add executables for chromedriver and geckodriver

HarlemSquirrel:
This adds executables similar to [`chromedriver-helper`](https://github.com/flavorjones/chromedriver-helper/tree/master/bin) and [`geckodriver-helper
`](https://github.com/DevicoSolutions/geckodriver-helper/tree/master/bin)

```sh
➤  bin/webdrivers-chromedriver --version                                                                                                                                          130 ↵
2019-05-14 16:32:54 WARN Webdrivers Driver caching is turned off in this version, but will be enabled by default in 4.x. Set the value with `Webdrivers#cache_time=` in seconds
ChromeDriver 74.0.3729.6 (255758eccf3d244491b8a1317aa76e1ce10d57e9-refs/branch-heads/3729@{#29})

➤  bin/webdrivers-geckodriver --version 
2019-05-14 16:33:03 WARN Webdrivers Driver caching is turned off in this version, but will be enabled by default in 4.x. Set the value with `Webdrivers#cache_time=` in seconds
geckodriver 0.24.0 ( 2019-01-28)

The source code of this program is available from
testing/geckodriver in https://hg.mozilla.org/mozilla-central.

This program is subject to the terms of the Mozilla Public License 2.0.
You can obtain a copy of the license at https://mozilla.org/MPL/2.0/.
```

### DIFF
--- a/bin/webdrivers-chromedriver
+++ b/bin/webdrivers-chromedriver
@@ -1,0 +1,6 @@
+#! /usr/bin/env ruby
+
+require_relative '../lib/webdrivers/chromedriver'
+
+Webdrivers::Chromedriver.update 
+exec Webdrivers::Chromedriver.driver_path, *ARGV
--- a/bin/webdrivers-chromedriver
+++ b/bin/webdrivers-chromedriver
@@ -1,6 +1,6 @@
 #! /usr/bin/env ruby
+# frozen_string_literal: true
 
 require_relative '../lib/webdrivers/chromedriver'
 
-Webdrivers::Chromedriver.update 
-exec Webdrivers::Chromedriver.driver_path, *ARGV
+Webdrivers::Chromedriver.run(*ARGV)
--- a/bin/webdrivers-geckodriver
+++ b/bin/webdrivers-geckodriver
@@ -1,0 +1,6 @@
+#! /usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative '../lib/webdrivers/geckodriver'
+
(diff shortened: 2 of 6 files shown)


twalpole:
I'm wondering what the use case for these is (especially since they wouldn't be using any settings made in an application that installed them)?

HarlemSquirrel:
For me, mostly a nice way to check the running version of the driver.

Kevin McCormack
Full Stack Web Developer
HarlemSquirrel.github.io

twalpole:
Ok, but since it wouldn't actually be loading things like the required version, or install dir, specified in an app, and if you haven't specified a required version webdrivers would be automatically updating it when the tests are run - does it actually provide useful info?

HarlemSquirrel:
I'm currently using something similar with `chromedriver-helper` and `geckodriver-helper` to display the version in use on CircleCI to assist in debugging. I see you're point so perhaps for this a better approach might be to print the version loaded by calling `Webdrivers::Chromedriver.desired_version`.

kapoorlakshya:
> might be to print the version loaded by calling Webdrivers::Chromedriver.desired_version.

@HarlemSquirrel `Webdrivers::Chromedriver.current_version` will give you that.

titusfortner:
Would it make sense to add a rake task `webdrivers:chromedriver:version` that outputs it?

HarlemSquirrel:
Rake tasks to display versions should work fine for my use case. If no one else sees a use for these executable additions then we can close this.

titusfortner:
Merged with #117

PR #121, opened 2019-05-17
Disable warning about caching in 4.x when setting `Webdrivers.cache_t…

ecbypi:
Hi there!

I recently updated to webdrivers v3.9.2 and found that setting `Webdrivers.cache_time` as mentioned in the warning about caching doesn't remove the warning.  This was kind of confusing and could be a problem on teams that are fastidious about ensuring there are no warnings or deprecations when running tests.  

I looked through the tests and didn't see anywhere that it made sense to add test coverage.  I see the behavior around `cache_time` is checked, but nothing for `cache_warning`.  Let me know if test coverage is a must for PRs.  

Thanks for the work on this gem.

### DIFF
--- a/lib/webdrivers/common.rb
+++ b/lib/webdrivers/common.rb
@@ -20,13 +20,20 @@ module Webdrivers
   class << self
     attr_accessor :proxy_addr, :proxy_port, :proxy_user, :proxy_pass, :install_dir
 
-    attr_writer :cache_time
-
     #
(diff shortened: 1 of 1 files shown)


ecbypi:
@titusfortner thanks!

titusfortner:
Oh, my comment didn't go through. This is better than what I had written, thanks for the PR. Just released 3.9.3 which includes this.

Issue #135, opened 2019-06-14
Provide (beginner friendly) advice on config files


matt17r:
#### Summary
It took me a while to figure out where to put config directives (i.e. `Webdrivers.cache_time` to prevent that warning). I'm fairly new to System Tests and I still have lots to learn about Rails in general but I think a minor tweak to the README would have helped me immensely.

#### Debug Info
At first I created a `webdrivers.rb` initialiser. It worked but I realised it will also be loaded in Prod and Dev, not just Test.

I then moved it to `config/environments/test.rb`. That worked too but I realised I only really need it when I run System Tests (which Rails doesn't do all the time by default).

Finally I moved it into `test/application_system_test_case.rb` which makes the most sense to me (but still might not be ideal).

#### Expected Behavior
Helpful information near the top of the README. Perhaps with examples for the most common test frameworks and then some general advice after that.

#### Actual Behavior
There is no info about this in the README.

matt17r:
I had a go at this myself in PR https://github.com/titusfortner/webdrivers/pull/136 but I'm not 100% confident in the advice I gave.

titusfortner:
Thanks for this and for pointing me to that article, there are a couple things that I need to comment on there.

Realize that not everyone using webdrivers is also using rails, so I can't speak to any rails specific needs for initializing, but it should go in whatever file gets required before executing tests, regardless of your environment.

@twalpole is there anything rails specific that is needed here? Should we suggest spec_helper or test_helper files?

twalpole:
I don't really think anything is needed here. Realistically anyone who is actually needing to add any webdrivers configuration (the cache warning should be gone if you're using the latest release) should already understand where test configurations belong in their specific project.

matt17r:
Respectfully, I disagree. Even a beginner may need to set up a proxy or want to adjust the logger level, both of which require configuring Webdrivers.

I now realise my instructions were a little bit too specific (given how many different ways this gem might get used) but @titusfortner said it much better than I did:
> it should go in whatever file gets required before executing tests

Those eleven words in the README won't impact those who should alredy understand but I think they would have pointed me in the right direction and saved me 10-15 minutes of ~Googling~ DuckDuckGo-ing and experimentation.

titusfortner:
Yes, but this is a "how does Ruby work" question, not a "how does webdrivers work" question. Hopefully your searching has given you more context for everything you will be doing going forward.

PR #136, opened 2019-06-14
Add recommended config locations to README

matt17r:
Add recommended locations for Webdriver configuration directives to the README for RSpec Specs and Rails (5.1+) System Tests

As per issue https://github.com/titusfortner/webdrivers/issues/135

I might be missing some important information. I blindly copied the RSpec location from https://bloggie.io/@kinopyo/migrate-from-chromedriver-helper-to-webdrivers

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,14 @@ require 'webdrivers'
 ```
 
 The drivers will now be automatically downloaded or updated when you launch a browser
-through Selenium. 
+through Selenium.
+
+### Configuration
+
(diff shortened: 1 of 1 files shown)


titusfortner (commented):
So, I'm not sure why your config values wouldn't just go in test helper or spec helper files, and I would hope that anyone running tests would know what file(s) get required during initialization.

I appreciate and applaud the desire to contribute, but I don't think this makes sense to add to the readme

Issue #130, opened 2019-06-07
Chrome binary cannot be located on Windows (JRuby)


joankaradimov:
#### Summary
The binary for Chrome cannot be found on Windows. It looks like it might be because of a whitespace in its path.

#### Debug Info
* Operating system: Windows 10
* Browser and version: Chrome 75
* Ruby: JRuby 9.2.7.0

#### Expected Behavior
No exception is thrown.

#### Actual Behavior
This exception is thrown:
```
x86 : The term 'x86' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:37
+ (Get-ItemProperty C:\Program Files (x86)\\Google\Chrome\Application\c ...
+                                     ~~~
    + CategoryInfo          : ObjectNotFound: (x86:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException


An error occurred in a `before(:suite)` hook.
Failure/Error: raise VersionError, 'Failed to find Chrome binary or its version.' if version.nil? || version.empty?

Webdrivers::VersionError:
  Failed to find Chrome binary or its version.
# ./features/build/tmp/jrubyExec/gems/webdrivers-4.0.0/lib/webdrivers/chrome_finder.rb:10:in `version'
# ./features/build/tmp/jrubyExec/gems/webdrivers-4.0.0/lib/webdrivers/chromedriver.rb:46:in `chrome_version'
# ./features/build/tmp/jrubyExec/gems/webdrivers-4.0.0/lib/webdrivers/chromedriver.rb:105:in `release_version'
# ./features/build/tmp/jrubyExec/gems/webdrivers-4.0.0/lib/webdrivers/chromedriver.rb:111:in `sufficient_binary?'
# ./features/build/tmp/jrubyExec/gems/webdrivers-4.0.0/lib/webdrivers/common.rb:132:in `correct_binary?'
# ./features/build/tmp/jrubyExec/gems/webdrivers-4.0.0/lib/webdrivers/common.rb:83:in `update'
# ./features/build/tmp/jrubyExec/gems/webdrivers-4.0.0/lib/webdrivers/chromedriver.rb:118:in `block in <main>'
# ./features/build/tmp/jrubyExec/gems/selenium-webdriver-3.142.3/lib/selenium/webdriver/common/service.rb:133:in `binary_path'
[...]

PR #138, opened 2019-06-17
Add support for WD_CHROME_PATH.

### LINKED ISSUE #137: Can we have a WD_CHROME_PATH ENV variable that sets chrome binary path?
#### Summary
Set Selenium::WebDriver::Chrome.path to ENV['WD_CHROME_PATH'] if it's defined.
This is similar to the WD_INSTALL_DIR env var for Webdrivers.install_dir

#### Debug Info
Please provide the following information for bug reports:

* Operating system / CI Environment:
* Browser and version:

#### Expected Behavior
If WD_CHROME_PATH is defined, then use that for Selenium::WebDriver::Chrome.path
Otherwise, use the default location.

#### Actual Behavior
Currently I have to muck around my spec_helper.rb but it would be nice to have this as part of webdriver itself to be consistent with WD_INSTALL_DIR

### LINKED ISSUE #139: Fails to start chrome on Mac because the filename contains spaces
#### Summary
When starting chrome through WebDrivers, WebDrivers asks the chrome binary for its version. This triggers an error as the google chrome binary contains spaces in its path on the Mac.

e.g. this will fail: 

```ruby
 `/Applications/Google Chrome.app/Contents/MacOS/Google Chrome --version`
```

To my understanding the problem is the in [system.rb:150](https://github.com/titusfortner/webdrivers/blob/master/lib/webdrivers/system.rb#L150) or in [chrome_finder.rb:71](https://github.com/titusfortner/webdrivers/blob/master/lib/webdrivers/chrome_finder.rb#L71) as it does not handle spaces in the path.

#### Debug Info
Please provide the following information for bug reports:

* Operating system / CI Environment: Mac OS 10.14.5
* Browser and version: Chrome Version 75.0.3770.90 (Offizieller Build) (64-Bit)

#### Expected Behavior
Space support in paths is required on Mac.

#### Actual Behavior
An error like

`*** Errno::ENOENT Exception: No such file or directory - /Users/dwt/Applications/Network/Browser/Google` is raised

kapoorlakshya:
* Adds support for `WD_CHROME_PATH` environment variable. ~~`WD_CHROME_PATH` takes precedence over `Selenium::WebDriver::Chrome.path`~~. Closes #137.

**Other changes**:

* ~~`WD_INSTALL_DIR` now has the highest precedence when determining the driver install (download) location. This is to make the behavior consistent with `WD_CACHE_TIME` and `WD_CHROME_PATH`.~~
* Removed ENV handing logic from the rake tasks. They now rely on the core code to handle these.

**UPDATE**:

* Fix bug where ENV vars took precedence over user defined configs in their project. The new order is:

1. User defined config
2. User defined ENV var value
3. Default value

* Fixes #139 and all other file path escaping issues by switching to `IO.popen`, which has OS specific path escaping logic built-in. Even handles JRuby on Windows specific issue #130.

### DIFF
--- a/spec/webdrivers/chrome_finder_spec.rb
+++ b/spec/webdrivers/chrome_finder_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Webdrivers::ChromeFinder do
+  let(:chrome_finder) { described_class }
+
+  context 'when the user relies on the gem to figure out the location of Chrome' do
+    it 'determines the location correctly based on the current OS' do
+      expect(chrome_finder.location).not_to be_nil
+    end
+  end
+
+  context 'when the user provides a path to the Chrome binary' do
+    it 'uses Selenium::WebDriver::Chrome.path when it is defined' do
+      Selenium::WebDriver::Chrome.path = chrome_finder.location
+      allow(chrome_finder).to receive(:win_location)
+      allow(chrome_finder).to receive(:mac_location)
+      allow(chrome_finder).to receive(:linux_location)
+      expect(chrome_finder.version).not_to be_nil
+      expect(chrome_finder).not_to have_received(:win_location)
+      expect(chrome_finder).not_to have_received(:mac_location)
+      expect(chrome_finder).not_to have_received(:linux_location)
+    end
+
+    it "uses ENV['WD_CHROME_PATH'] when it is defined" do
+      allow(ENV).to receive(:[]).with('WD_CHROME_PATH').and_return(chrome_finder.location)
+      allow(chrome_finder).to receive(:win_location)
+      allow(chrome_finder).to receive(:mac_location)
+      allow(chrome_finder).to receive(:linux_location)
+      expect(chrome_finder.version).not_to be_nil
+      expect(chrome_finder).not_to have_received(:win_location)
+      expect(chrome_finder).not_to have_received(:mac_location)
+      expect(chrome_finder).not_to have_received(:linux_location)
+    end
+
+    it 'uses Selenium::WebDriver::Chrome.path over WD_CHROME_PATH' do
+      Selenium::WebDriver::Chrome.path = chrome_finder.location
+      allow(ENV).to receive(:[]).with('WD_CHROME_PATH').and_return('my_wd_chrome_path')
+      expect(chrome_finder.version).not_to be_nil
+      expect(ENV).not_to have_received(:[]).with('WD_CHROME_PATH')
+    end
+  end
+end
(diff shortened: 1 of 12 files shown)


twalpole:
As implemented ENV['WD_CHROME_PATH'] is preferred over Selenium::WebDriver::Chrome.path (and the equivalents for other drivers)- I think that should probably be the other way around since code setting is more deliberate than environment variable.

kapoorlakshya:
As discussed on Slack, the current behavior of ENV taking precedence over code is now categorized as a bug. I'll update the PR tonight with the new order: 

1. User defined config
2. User defined ENV var
3. Default value

This way any deliberate changes in the code will take precedence over ENV values.

kapoorlakshya:
@twalpole PR is ready for second review.

spec/webdrivers/chrome_finder_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  context 'when the user provides a path to the Chrome binary' do
+    it 'uses Selenium::WebDriver::Chrome.path when it is defined' do
+      Selenium::WebDriver::Chrome.path = chrome_finder.send("#{platform}_location")
+      allow(chrome_finder).to receive(:win_location)
+      allow(chrome_finder).to receive(:mac_location)
+      allow(chrome_finder).to receive(:linux_location)
+      chrome_finder.version
```
twalpole: should probably verify this isn't nil

spec/webdrivers/chrome_finder_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect(chrome_finder).not_to have_received(:win_location)
+      expect(chrome_finder).not_to have_received(:mac_location)
+      expect(chrome_finder).not_to have_received(:linux_location)
+    end
+
+    it "uses ENV['WD_CHROME_PATH'] when it is defined" do
+      loc_method = "#{platform}_location"
+      allow(ENV).to receive(:[]).with('WD_CHROME_PATH').and_return(chrome_finder.send(loc_method))
```
twalpole: maybe we should add a `location` method to the finder classes that does the calling of platform specific versions

twalpole (approved):
Looks good to me -- other than possibly adding `location` method to finder classes

spec/webdrivers/chrome_finder_spec.rb
kapoorlakshya: I added `ChromeFinder.location` and made the platform specific location and version methods `private`. Let me know if you think we should keep them public for any reason.

spec/webdrivers/chrome_finder_spec.rb
kapoorlakshya: Done.

twalpole:
That looks better - 👍

PR #152, opened 2019-09-27
add parallel_tests gem documentation

griffincox:
Adds documentation for use with the parallel_tests gem


kapoorlakshya:
Hi @griffincox, thanks for your contribution. I think this information would better fit in the [wiki](https://github.com/titusfortner/webdrivers/wiki) section. Can you please create a new page there instead?

The README file already has a lot of information in it.

kapoorlakshya:
I have created the wiki page here - https://github.com/titusfortner/webdrivers/wiki/parallel_tests-gem

Thanks again for your contribution!

PR #153, opened 2019-09-30
Upgrade rubyzip to 1.3.0 for known vulnerability

rhymes:
This updates the minimum required version of rubyzip due to a known vulnerability
https://github.com/rubyzip/rubyzip/pull/403

Due to the fact that this gem downloads zip from known and trusted sources it's probably not an urgent update.

### DIFF
--- a/lib/webdrivers/common.rb
+++ b/lib/webdrivers/common.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'rubygems/package'
-require 'zip'
 require 'webdrivers/logger'
 require 'webdrivers/network'
 require 'webdrivers/system'
--- a/lib/webdrivers/system.rb
+++ b/lib/webdrivers/system.rb
@@ -4,6 +4,12 @@ require 'rubygems/package'
 require 'zip'
 require 'English'
 
+# validate zip entry sizes to avoid zip bombs
+# see https://github.com/rubyzip/rubyzip#size-validation
+# and https://github.com/rubyzip/rubyzip/pull/403 for further details
+# this will be the default in rubyzip 2.0+
+Zip.validate_entry_sizes = true
+
 module Webdrivers
   #
   # @api private
--- a/webdrivers.gemspec
+++ b/webdrivers.gemspec
@@ -29,6 +29,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'simplecov', '~>0.16'
 
   s.add_runtime_dependency 'nokogiri', '~> 1.6'
-  s.add_runtime_dependency 'rubyzip', '>= 1.2.2'
+  s.add_runtime_dependency 'rubyzip', '>= 1.3.0'
   s.add_runtime_dependency 'selenium-webdriver', '>= 3.0', '< 4.0'
 end
--- a/webdrivers.gemspec
+++ b/webdrivers.gemspec
@@ -29,6 +29,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'simplecov', '~>0.16'
 
   s.add_runtime_dependency 'nokogiri', '~> 1.6'
-  s.add_runtime_dependency 'rubyzip', '~> 1.0'
+  s.add_runtime_dependency 'rubyzip', '~> 1.0', '>= 1.3.0'
   s.add_runtime_dependency 'selenium-webdriver', '>= 3.0', '< 4.0'
 end
--- a/webdrivers.gemspec
+++ b/webdrivers.gemspec
@@ -29,6 +29,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'simplecov', '~>0.16'
 
   s.add_runtime_dependency 'nokogiri', '~> 1.6'
-  s.add_runtime_dependency 'rubyzip', '~> 1.0', '>= 1.3.0'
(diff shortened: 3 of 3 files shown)


rhymes:
BTW rubyzip has also released 2.0 with the flag enabled, let me know if you prefer a more elaborate PR that takes that into account as well

rwojnarowski:
Selenium relaxed rubyzip version to allow 2.0 (`'>= 1.2.2'`), https://github.com/SeleniumHQ/selenium/pull/7613/files. I think it's a good idea to do same thing here as well

kapoorlakshya:
@rhymes Thanks for the PR! I agree with @rwojnarowski. Can you please change the version requirement to be `'>= 1.2.2'`?

woodhull:
Yes, it seems important to merge this to allow downstream projects to start using rubyzip 2.0

rhymes:
@kapoorlakshya I relaxed the version but then I'm thinking we might need to check which version is running for the flag here https://github.com/titusfortner/webdrivers/pull/153/files#diff-148937c344e3c9bb067f1360bd9d14b2R7

because if it's below `1.3.0` that flag doesn't exist

atsheehan:
@rhymes do we need to set the global flag on `Zip`? It might be confusing if an app wants to keep it at false and this gem overrides it. 

Maybe a note in the README recommending users upgrade to rubyzip 2.0, or adding the option to the `extract` method call?

rhymes:
@atsheehan according to the PR https://github.com/rubyzip/rubyzip/pull/403 - it needs to be set if one uses a version `rubyzip >= 1.3.0 and < 2.0.0`, if you use a version earlier than 1.3.0 you have the vulnerability, if you use 2.0 the default of that global value is already true.

I'm not sure why we want to allow versions under `1.3.0` TBH but I'm okay either way.

I think we should check the version at runtime and act accordingly in relation to the global flag

twalpole:
I agree there’s no reason to allow < 1.3 on a new release of webdrivers

anilreddy:
@kapoorlakshya can you please merge it and make release as we are facing this issue while using webdrivers.

Error:
------
```ruby
Unable to activate webdrivers-4.1.2, because rubyzip-2.0.0 conflicts with rubyzip (~> 1.0) (Gem::ConflictError)
C:/Ruby26-x64/lib/ruby/2.6.0/rubygems/specification.rb:2302:in `raise_if_conflicts'
C:/Ruby26-x64/lib/ruby/2.6.0/rubygems/specification.rb:1418:in `activate'
C:/Ruby26-x64/lib/ruby/2.6.0/rubygems.rb:223:in `rescue in try_activate'
C:/Ruby26-x64/lib/ruby/2.6.0/rubygems.rb:216:in `try_activate'
C:/Ruby26-x64/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:123:in `rescue in require'                             C:/Ruby26-x64/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:34:in `require'
F:/Watir/features/support/env.rb:9:in `<top (required)>'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/glue/registry_and_more.rb:107:in `load'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/glue/registry_and_more.rb:107:in `load_code_file'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/runtime/support_code.rb:144:in `load_file'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/runtime/support_code.rb:85:in `block in load_files!'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/runtime/support_code.rb:84:in `each'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/runtime/support_code.rb:84:in `load_files!'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/runtime.rb:272:in `load_step_definitions'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/runtime.rb:68:in `run!'
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/lib/cucumber/cli/main.rb:34:in `execute!'                         C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/cucumber-3.1.2/bin/cucumber:9:in `<top (required)>'
C:/Ruby26-x64/bin/cucumber:23:in `load'
C:/Ruby26-x64/bin/cucumber:23:in `<main>'
```

kapoorlakshya:
@rhymes Let's go ahead and change it to back `>= 1.3.0` to make sure a vulnerable version is never used. Thank you!

**Edit**: I just realized you allowed us to edit the PR so I have made the change myself.

@anilreddy Yes, will release as soon as the requested change is made.

kapoorlakshya:
Released v4.1.3. Let me know if anything breaks.

rhymes:
@kapoorlakshya thanks for taking it to the finish line!

PR #172, opened 2020-05-23
Implement initial support for WSLv1

### LINKED ISSUE #170: Support downloading exe when running WSL
#### Summary
I use WSL for development, which works really well for most things, but one of the last hurdles I've been trying see if I can nicely resolve is around feature testing with Chrome via Selenium.

Currently this can be accomplished by using `chromedriver.exe`, and having Selenium use that instead of letting it try and start chromedriver itself (aka "don't run webdriver, we'll run webdriver")

The problem is that when using `webdrivers`, it sees itself as being on Linux (understandably because it mostly is), and attempts to do all the Linux things, such as looking for Chrome in the Linux places, and downloading the Linux version of the webdriver.

Initially what happens is that `webdrivers` attempts to find chrome, and fails, throwing an error (understandably) - to work around this, I can set `WD_CHROME_PATH` to point to a bash script (aka `fake-chrome.sh`).

This makes `webdrivers` happy, and I just have to run `chromedriver.exe` myself in a terminal, and set the `url:` property.

However, it's actually possible to improve this further so that `webdrivers` does its thing on WSL as it does for native Windows, Linux, and MacOS - that is, it downloading the right version of the webdriver based on the chrome version.

Since WSL supports running exes, and `chromedriver.exe` supports flags in both Windows & Linux format (so both `chromedriver.exe --version` & `chromedriver.exe /version` works), you can actually replace the Linux `chromedriver` downloaded by `webdrivers` with the Windows version, and things will "just work".

This leads into the actual feature request of this issue: `webdrivers` on WSL will download the Linux version currently.

If `webdrivers` supported downloading the Windows version of chromedriver, named as just `chromedriver` (so that the rest of the code didn't need to change), then WSL support would be nearly on par with the rest.

You can check if you're running on WSL a few different ways, but there doesn't seem to be a gem version of [`is-wsl`](https://github.com/sindresorhus/is-wsl/blob/master/index.js) - I'm personally happy for this to be walled off behind an explicit env value, since you'll still need to point to a `fake-chrome` script file.

The contents of that file at this point would contain the following:

```
#!/usr/bin/env bash

echo $(powershell.exe '(Get-ItemProperty "C:\Program Files (x86)\Google\Chrome\Application\chrome.exe").VersionInfo.ProductVersion')
```

-----------

My apologies for this rather lengthy issue, but I wanted to lay out the whole story.

I'm happy to do the actual implementation work, but wanted to check if this is something `webdrivers` would be happy to have in the first place :)

Ultimately the first iteration of an implementation I'd gun for is "if WD_IN_WSL: download driver exe and rename, then carry on as if linux".

G-Rath:
This implementation is working for me locally without any edge cases, so wanted to open a PR for initial feedback on the structure.

I found during my implementing that it's fine for `chromedriver` to be called `chromedriver.exe`, which meant I didn't need to add any extra logic around extracting the file from the zip & then knowing to rename it.

I've not looked into the tests yet: unfortunately it's not really possible to write tests for WSL right now, as enabling WSL requires a restart which CIs & containers either don't support or crash on, but if you're ok with mocking/spying I can write some tests to make sure the right methods are called (which is how I've done it in other libraries).

closes #170

### DIFF
--- a/lib/webdrivers/chrome_finder.rb
+++ b/lib/webdrivers/chrome_finder.rb
@@ -50,6 +50,28 @@ module Webdrivers
         nil
       end
 
+      def wsl_location
+        _, drive, user = ENV['PATH'].match(%r{/([a-z])/Users/([^/:]+)/AppData/}).to_a
+
+        roots = [
+          "#{drive}:\\Users\\#{user}\\AppData\\Local",
+          "#{drive}:\\Program Files (x86)",
+          "#{drive}:\\Program Files"
+        ]
+
+        directories = %w[Google\\Chrome\\Application Chromium\\Application]
+        file = 'chrome.exe'
+
+        directories.each do |dir|
+          roots.each do |root|
+            option = System.to_wsl_path("#{root}\\#{dir}\\#{file}")
+            return option if File.exist?(option)
+          end
+        end
+
+        nil
+      end
+
       def mac_location
         directories = ['', File.expand_path('~')]
         files = ['/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
@@ -66,6 +88,8 @@ module Webdrivers
       end
 
       def linux_location
+        return wsl_location if System.wsl?
+
         directories = %w[/usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /snap/bin /opt/google/chrome]
         files = %w[google-chrome chrome chromium chromium-browser]
 
@@ -80,10 +104,16 @@ module Webdrivers
       end
 
       def win_version(location)
-        System.call("powershell (Get-ItemProperty '#{location}').VersionInfo.ProductVersion")&.strip
+        System.call("powershell.exe \"(Get-ItemProperty '#{location}').VersionInfo.ProductVersion\"")&.strip
+      end
+
+      def wsl_version(location)
+        win_version(System.to_win32_path(location))
       end
 
       def linux_version(location)
+        return wsl_version(location) if System.wsl?
+
         System.call(location, '--product-version')&.strip
       end
 
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -80,7 +80,7 @@ module Webdrivers
       end
 
       def file_name
-        System.platform == 'win' ? 'chromedriver.exe' : 'chromedriver'
+        System.platform == 'win' || System.wsl? ? 'chromedriver.exe' : 'chromedriver'
       end
 
       def download_url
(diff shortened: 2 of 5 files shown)


kapoorlakshya:
> but if you're ok with mocking/spying I can write some tests to make sure the right methods are called

@G-Rath That would be great! Thank you!

lukepearcewp:
This is great! I was just having a load of trouble trying to get rspec playing nicely on a colleagues windows WSL setup and this just works :)

The only odd thing I did notice is that javascript console.log() messages are output to stdout which you don't normally see whilst running tests - not sure if this is avoidable?

kapoorlakshya:
@G-Rath Thanks again for all the work! I'll try to review and hopefully merge it this week.

kapoorlakshya:
@G-Rath The code looks good and the specs pass for me locally on WSL v1 (Windows 10 build 1903). Thank you again for your contribution! :)

P.S. I have created #173 to configure Appveyor (CI) with WSL to test the support long term. Please feel free to submit a PR for that if you're interested. Otherwise I'll get to it when I can.

G-Rath:
🥳 awesome! Happy to help - feel free to ping me on any WSL stuff that comes up (even if it's not a result of this PR).

> Please feel free to submit a PR for that if you're interested. Otherwise I'll get to it when I can.

Oh wow that's cool! From everything I've read so far (admittedly not a lot outside of a few docker & github-actions feature-request-issues) it didn't seem like there was a CI that supported WSL testing as it requires a restart of the container to successfully complete setup which the CI host usually just says "must be finished" and prevents it from starting back up.

I've not used Appveyor before, but super keen to have a go at setting it up, since it could be used by a few of the other projects I've recently been implementing WSL support in.

Cheers!

G-Rath:
Just remembered I meant to mention this but didn't: I'm pretty sure this logic should mostly "just work" for at least the Edge driver too, but haven't got it installed locally & we don't use it at work, so at some point when I've got the time I'll confirm if that's the case & open an issue or PR depending on how far I get.

(if the Appveyor setup works out, I'll probably do it after that as it'll make it easier anyway).

kapoorlakshya:
@G-Rath Sounds good! Looking forward to more contributions from you.

G-Rath:
> The only odd thing I did notice is that javascript console.log() messages are output to stdout which you don't normally see whilst running tests - not sure if this is avoidable?

@lukepearcewp so this is interesting: that doesn't happen when running in tmux, which is why I didn't notice it the first time.

So..., run it in tmux apparently is the nice answer?

I imagine that there is probably a `--silent` flag that it can be called with, which I'll look into when I've got a chance.

-------

This also fixes an annoying terminal font change that happens with [powershell](https://github.com/microsoft/terminal/issues/367), which I'm looking into ways around - I might make a PR to do something like trying to call the new powershell and if that doesn't exist use the old one, but it's good to know that tmux solves both those problems as there's no downside to running it in a tmux panel anyway 🤷

lukepearcewp:
> So..., run it in tmux apparently is the nice answer?

Ha, if it was my setup I wouldn't have noticed it either then ;)

Yeah it's not really a problem from my point of view, just something that stuck out as odd.  I would look into it myself but I'm on a mac - trying to diagnose and fix stuff over a screen-sharing session on someone else's Windows setup isn't that fun :)

Anyways thanks again @G-Rath saved a load of headaches with this pull request.

PR #176, opened 2020-06-07
Support using new powershell if available

G-Rath:
It seems that `tmux` does not always prevent the console from being made weird 😬

There's a bit of a spider web of github issues, but the main takeaway is that while MS would love to fix it, it's very unlikely they'll be able to without breaking a lot of other things, and ideally just use the new powershell.

I've included a link to [one of the issues](https://github.com/microsoft/terminal/issues/367) that serve as an entry into the web for those wanting more details :)

You can reproduce the underlying issue by just running `powershell.exe | true` on WSL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,14 @@ The logging level can be configured for debugging purpose:
 Webdrivers.logger.level = :DEBUG
 ```
 
+### WSLv1 support
+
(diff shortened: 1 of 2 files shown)


kapoorlakshya:
@G-Rath Thanks again for your contribution!

PR #183, opened 2020-09-16
Gemfile: require: false

runephilosof-abtion:
Otherwise the requires below in the documentation doesn't make sense.


kapoorlakshya:
@runephilosof-abtion Thank you for your contribution!

Ignoring CI failures and merging since this is a readme-only change.

PR #193, opened 2021-01-07
Add support for Apple M1 architecture (chromedriver)

### LINKED ISSUE #191: Google Chrome on Apple M1 architecture
Maybe I'm not at the right place to submit this issue. Feel free to close and redirect me.

I have a new Macbook M1, and I installed the M1-specific Google Chrome on my system to avoid using the Rosetta emulation. It works well!

However, when I'm running my test suite, I can see in the Activity Monitor that the headless Chrome is not started with the correct architecture. We can see "Intel" instead of "Apple":

Screenshot 2020-12-30 at 16 32 37

I checked and Google Chrome seems to use Universal Binary, so either Intel or Apple, but the wrong one is started.

Would it be because chromedriver doesn't officially support M1 architecture? [1] And Google Chrome is forced to follow the same architecture? 

Is there any way to fix this situation in webdrivers, and force Google Chrome with M1 architecture? It will certainly speed up significantly my test suite.

Or maybe I need to wait when (if?) a M1-specific chromedriver is released?

Any thoughts on this  would be welcome 😄 

[1] https://chromedriver.storage.googleapis.com/index.html?path=87.0.4280.88/

MichaelHoste:
Fixes #191

I hope it's the correct approach.

It's still compatible with Chromedriver versions lower than `87.0.4280.88` but it will then fallback to the Intel version of Chrome.

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -87,6 +87,14 @@ module Webdrivers
         System.platform == 'win' || System.wsl? ? 'chromedriver.exe' : 'chromedriver'
       end
 
+      def apple_m1_cpu?
+        RUBY_PLATFORM.include?('arm64-darwin')
+      end
+
+      def apple_m1_compatible?(version)
+        version >= normalize_version('87.0.4280.88')
+      end
+
       def download_url
         return @download_url if @download_url
 
@@ -96,7 +104,9 @@ module Webdrivers
                     normalize_version(required_version)
                   end
 
-        file_name = System.platform == 'win' || System.wsl? ? 'win32' : "#{System.platform}64"
+        archi = apple_m1_cpu? && apple_m1_compatible?(version) ? '_m1' : ''
+
+        file_name = System.platform == 'win' || System.wsl? ? 'win32' : "#{System.platform}64#{archi}"
         url = "#{base_url}/#{version}/chromedriver_#{file_name}.zip"
         Webdrivers.logger.debug "chromedriver URL: #{url}"
         @download_url = url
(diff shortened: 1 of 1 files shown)


lib/webdrivers/chromedriver.rb
```diff
@@ -87,6 +87,14 @@ def file_name
         System.platform == 'win' || System.wsl? ? 'chromedriver.exe' : 'chromedriver'
       end
 
+      def apple_m1_cpu?
+        RUBY_PLATFORM.include?('arm64-darwin')
+      end
+
+      def apple_m1_compatible?(version)
```
kapoorlakshya, line 90: How about we combine `apple_m1_cpu?` and `apple_m1_compatible ?` as one `apple_m1_compatible?`? Then line 107 can simply be:

```ruby
apple_arch = apple_m1_compatible?(version) ? '_m1' : ''
```

I am not seeing a use case where we would ever use these methods separately. Do you?

lib/webdrivers/chromedriver.rb
```diff
@@ -96,7 +104,9 @@ def download_url
                     normalize_version(required_version)
                   end
 
-        file_name = System.platform == 'win' || System.wsl? ? 'win32' : "#{System.platform}64"
+        archi = apple_m1_cpu? && apple_m1_compatible?(version) ? '_m1' : ''
```
kapoorlakshya: Can we please rename `archi` to `apple_arch` or `mac_arch` since it's specific to macOS/apple? Another option is to pull out the filename logic into a separate method to keep the complexity low.

kapoorlakshya (changes requested):
@MichaelHoste Thank you very much for the PR! Left two comments, but the overall approach looks good to me.

MichaelHoste:
I did what you suggested, it made sense!

I first tried to create a separate method for the storage file name, but since we need to pass `version` with it, it was becoming too verbose and confusing for something that was already quite easy to read and understand. I'm not sure it was better.

kapoorlakshya:
@MichaelHoste Thank you for making the changes. Looks good!

Issue #192, opened 2020-12-30
Can't find Chrome Binary in WSL2 with version 4.4.2


fabioxgn:
After updating to v4.4.2 my tests are failing with:

```
Webdrivers::BrowserNotFound:  Failed to find Chrome binary.
```

I've tried to set the `WD_CHROME_PATH` to the chrome binary path of my WSL2 Ubuntu installation but it doesn't work either:

```
 Webdrivers::VersionError: Failed to find Chrome version
```

```
 RuntimeError:
            Failed to make system call: ["/home/fabio/.webdrivers/chromedriver.exe", "--version"]
```

What's the proper way to use chrome on WSL2?

kapoorlakshya:
Was this working in v4.4.1? And please share the debug logs ([instructions](https://github.com/titusfortner/webdrivers#logging)). I'll try to setup WSL2 tomorrow and reproduce this.

@G-Rath Would it be possible for you to troubleshoot this as well?

G-Rath:
@kapoorlakshya I don't have WSLv2 setup, but happy to help where I can - I'll see if I can find some time to try get it setup somewhere so I can test things locally.

kapoorlakshya:
@G-Rath Thanks! Here's what I've found:

v4.4.1 and v4.4.2 are able to find the Chrome binary installed on my local system when using WSL2:

```
C:\Users\Lakshya>wsl -l -v
  NAME            STATE           VERSION
* Ubuntu-20.04    Running         1

C:\Users\Lakshya>wsl
kapoorlakshya@Windows10:/mnt/c/Users/Lakshya$

kapoorlakshya@Windows10:/mnt/c/Users/Lakshya$ ruby -v
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]

kapoorlakshya@Windows10:/mnt/c/Users/Lakshya$ irb
irb(main):001:0> require 'webdrivers'
=> true

irb(main):003:0> Webdrivers::ChromeFinder.location
=> "/mnt/c/Program Files (x86)/Google/Chrome/Application/chrome.exe"

irb(main):004:0> Webdrivers::Chromedriver.browser_version
sh: 1: pwsh.exe: not found
=> #<Gem::Version "87.0.4280.88">

irb(main):005:0> Webdrivers::Chromedriver.current_version
=> nil

irb(main):006:0> Webdrivers::Chromedriver.update
sh: 1: pwsh.exe: not found
[...]

PR #195, opened 2021-01-08
Adds option to use chrome on Linux when running with WSL

fabioxgn:
As discussed in #192, after v4.4.2 it is impossible to use Chrome on Linux when running in WSL as the system is detected as WSL and tries to always use the Chrome on Windows.

I don't know if this is the best solution to this, but i'd like to have the option to run Chrome on Linux and this solves my issue. 

I'm not sure about the name of the env var and if this is clear enough, I'll be happy to accept suggestions.


fabioxgn:
I'm closing this because I think that this is unecessary. We should detected that it is running on WSL2 and use Chrome on Linux, as explained here: https://github.com/titusfortner/webdrivers/issues/192#issuecomment-757130364

Issue #197, opened 2021-01-11
Problem using the gem in docker inside WSL2


urubatan:
https://github.com/titusfortner/webdrivers/blob/7ef83985ee9edf16de7599db60aad0e6df7d54c0/lib/webdrivers/system.rb#L152

This test to check WSL is causing problems if the gem is used inside docker at least if the server is executed in WSL2

dirkkelly:
I am having the same issue, could we override this with an environment flag?

kapoorlakshya:
Closed by #196. Fix confirmed to be working by @urubatan.

PR #196, opened 2021-01-09
Does not use Chrome on Windows on WSLv2

fabioxgn:
Fixes #192 

This PR https://github.com/titusfortner/webdrivers/pull/187 changed WSL detection to detect both WSLv1 and WSLv2 as running in WSL and use Chrome on Windows for both versions. But it doesn't work on WSLv2 out of the box.

WSL2 networking is different from WSL1 and you can't connect from the WSL2 Linux on Chromedriver running on Windows using `127.0.0.1:9515`, see: https://github.com/microsoft/WSL/issues/4619

This PR changes the wsl detection to be explicity about v1 and only uses Chromedriver on Windows if it is WSLv1, as it doesn't work on WSLv2.

PR #187, opened 2020-12-17
Make WSL detection more generic

rbclark:
On WSL2, the contents of `/proc/version` seem to have changed slightly, the command currently in webdrivers currently returns false:

```ruby
irb(main):001:0> File.open('/proc/version').read.include?('Microsoft')
=> false
irb(main):002:0> File.open('/proc/version').read.downcase.include?('microsoft')
=> true
```

Full contents for completeness:

```ruby
irb(main):003:0> File.open('/proc/version').read
=> "Linux version 4.19.104-microsoft-standard (oe-user@oe-host) (gcc version 8.2.0 (GCC)) #1 SMP Wed Feb 19 06:37:35 UTC 2020\n"
```

Unfortunately this does not seem to fix all of the issues, although I am unsure if the rest of the issues are in the webdrivers gem or not. I am still receiving a `Selenium::WebDriver::Error::WebDriverError: unable to connect to chromedriver 127.0.0.1:9515` error when trying to run on WSL2 with a pretty barebones Rails system test setup.

PR #199, opened 2021-01-13
Change the way the library uses to detect WSL to detect only WSL v1

### LINKED ISSUE #192: Can't find Chrome Binary in WSL2 with version 4.4.2

urubatan:
Uses the kernel version to detect if the current kernel version is V1 or V2

This is the correct implementation for https://github.com/titusfortner/webdrivers/pull/196 and all the description there applies here

Fixes #192 and #197

This PR #187 changed WSL detection to detect both WSLv1 and WSLv2 as running in WSL and use Chrome on Windows for both versions. But it doesn't work on WSLv2 out of the box.

WSL2 networking is different from WSL1 and you can't connect from the WSL2 Linux on Chromedriver running on Windows using 127.0.0.1:9515, see: microsoft/WSL#4619

This PR changes the wsl detection to be explicity about v1 and only uses Chromedriver on Windows if it is WSLv1, as it doesn't work on WSLv2.


urubatan:
@titusfortner I also added tests to check if it wont identify WSL2 by mistake

kapoorlakshya:
Closing at user's [request](https://github.com/titusfortner/webdrivers/pull/196#issuecomment-759733934).

PR #215, opened 2021-09-10
Linux edge fixes

luke-hill:
* Allow Linux Edge to be automatically detected and interrogated for browser version
* Allow msedgedriver to be automatically detected and downloaded on Linux
* Prevent NME caused by selenium versions not being fully updated

### DIFF
--- a/lib/webdrivers/edge_finder.rb
+++ b/lib/webdrivers/edge_finder.rb
@@ -24,6 +24,8 @@ module Webdrivers
       private
 
       def user_defined_location
+        return unless defined? Selenium::WebDriver::Edge.path
+
         if Selenium::WebDriver::Edge.path
           Webdrivers.logger.debug "Selenium::WebDriver::Edge.path: #{Selenium::WebDriver::Edge.path}"
           return Selenium::WebDriver::Edge.path
--- a/lib/webdrivers/edge_finder.rb
+++ b/lib/webdrivers/edge_finder.rb
@@ -69,18 +69,16 @@ module Webdrivers
       end
 
       def linux_location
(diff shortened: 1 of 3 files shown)


lib/webdrivers/edge_finder.rb
```diff
@@ -24,6 +24,8 @@ def location
       private
 
       def user_defined_location
+        return unless defined? Selenium::WebDriver::Edge.path
```
titusfortner, line 27: Why do you need this line? If it isn't defined it probably isn't going to do the right thing.

luke-hill: This permits people using this gem and slightly older versions of selenium. This way you don't get crashes, it just fails gracefully and then attempts to find the location using the System platform location methods.

titusfortner: The versions of Selenium that don't have this method don't support Chromium Edge, so it shouldn't matter. Do you have a specific scenario where an error was raised, but getting a nil would have done the right thing?

luke-hill: Still using 4alpha6. Think I've mentioned. This prevents a NME

titusfortner: Hah, I thought `#path` had been in there longer. It was added for alpha 7. You should be able to add this and have it work in alpha 6:
```
Selenium::WebDriver::Edge = Selenium::WebDriver::EdgeChrome
```

luke-hill: Yep, but is there any harm in adding this as a preventative measure, given it doesn't change what the function does.

titusfortner:
I uncommented @kapoorlakshya's code for 4.7
5.0 is going to require Selenium 4+ so the conditional won't be needed

luke-hill:
Awesome stuff thanks!

PR #221, opened 2021-10-18
Fix entry of v4.7.0

y-yagi:
It seems that this is Ruby's version.
https://github.com/titusfortner/webdrivers/commit/c209cf82944afbb156b741f21adba7d499470967


titusfortner:
Nice catch, thanks.

Issue #217, opened 2021-09-27
Use snap driver when chromium is installed via snap package


mdesantis:
#### Summary

The downloaded chromedriver doesn't work with Chromium installed via snap package (as written here: https://github.com/titusfortner/webdrivers/wiki/Chrome-snap-package-hangs).

The solution pointed in the wiki suggests installing Chromium via apt, but it isn't an easy option as Ubuntu doesn't provide an apt flavour of Chromium and the unofficial alternatives are unreliable (for example, the one linked in the issue related to the wiki page is marked as abandoned: https://launchpad.net/~xalt7x/+archive/ubuntu/chromium-deb-vaapi).

Anyway, The snap flavour of chromedriver works fine together with snap Chromium. So I suggest to detect that Chromium is installed via snap (I guess it is possible to check that the Chromium path starts with `/snap/`) and, in this case, skip chromedriver installation and use the snap chromedriver (or tell the user to install it). I'm open to work on a pull request, if welcome.

My current workaround is to overwrite the chromedriver path set by `webdrivers`:

```ruby
::Selenium::WebDriver::Chrome::Service.driver_path = proc { '/snap/bin/chromium.chromedriver' }
```

#### Debug Info

* Webdrivers version: 4.6.1
* Ruby version: 3.0.2
* Operating system / CI Environment: Linux Ubuntu 20.04.2
* Browser and version:
  - Chromium 93.0.4577.82 snap
  - ChromeDriver 93.0.4577.82 (e3a25d9b9e2d0b728e045ec87c0aa4942aa46e4e-refs/branch-heads/4577@{#1237})

#### Expected Behavior

I expect Webdrivers to detect that Chromium is installed via snap and to set the related chromedriver accordingly, rather than downloading and hanging during tests running.

#### Actual Behavior

Well, I guess I can write here the error log when using snap Chromium together with the downloaded driver:

<details>
<pre><code>
> RAILS_ENV=test bin/rails test:system
Running 4 tests in a single process (parallelization threshold is 50)
Run options: --seed 32101

# Running:

E

Error:
PagesTest#test_should_destroy_Page:
Selenium::WebDriver::Error::UnknownError: unknown error: DevToolsActivePort file doesn't exist
    test/system/pages_test.rb:40:in `block in <class:PagesTest>'

Error:
PagesTest#test_should_destroy_Page:
Net::ReadTimeout: Net::ReadTimeout with #<TCPSocket:(closed)>
[...]

PR #227, opened 2022-01-24
Set `WEBDRIVERS=1` env var while running drivers

entretechno-jeremiah:
See <https://github.com/titusfortner/webdrivers/issues/217#issuecomment-1019405659>

@kapoorlakshya I've tested this on my machine (Ubuntu 21.10), and I've confirmed that it does make the env var available to my script. If you would be able to test this on Windows (mainly to make sure that it didn't break anything, not so much that the env var is present), that would be amazing. Thank you!

The reason I added intermediate scripts in `bin/drivers/` is so that the env var is only set while that driver is actually running. I wanted to make sure that if a user launches a driver via this library and then launches a driver **not** via this library in the same script, it would have the env var set for the former but not the latter.

I totally understand if this is not an ideal change to this library. If not, my existing solution should work fine without the env var.

### DIFF
--- a/bin/drivers/chromedriver
+++ b/bin/drivers/chromedriver
@@ -1,0 +1,7 @@
+#!/usr/bin/env ruby
+
+require 'webdrivers/chromedriver'
+
+ENV['WEBDRIVERS'] = '1'
+
+exec ::Webdrivers::Chromedriver.update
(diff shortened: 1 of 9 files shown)


titusfortner:
Yeah, this seems like an unnecessary complexity for a non-standard use case. You should be able to implement this in an alternate fashion locally if you need it.

Thanks!

PR #228, opened 2022-01-26
Add Ruby 3.1 to CI

petergoldstein:
This PR adds Ruby 3.1 to the CI matrix

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         os:  [ubuntu-18.04, macos-10.15, windows-2019]
-        ruby-version: [2.6, 2.7, 3.0, jruby]
+        ruby-version: [2.6, 2.7, '3.0', 3.1, jruby]
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         os:  [ubuntu-18.04, macos-10.15, windows-2019]
-        ruby-version: [2.6, 2.7, 3.0, jruby]
+        ruby-version: [2.6, 2.7, '3.0', 3.1, jruby]
```
kapoorlakshya, line 30: Are the single quotes needed for some reason or should they be removed?

petergoldstein: Quotes are required around 3.0 because without quotes YAML truncates the value to 3.  That causes the latest Ruby 3 version to be loaded - at this time Ruby 3.1.0.

To ensure a Ruby 3.0.x is used for this matrix element, now and in the future, it must be quoted.

kapoorlakshya: Interesting... Didn't know that. Thanks!

PR #232, opened 2022-05-04
Improve error messages for debugging

dan-jensen:
This improves 2 error messages in ChromeFinder. Previously the messages
could imply that a file could not be found. It is more accurate to
say the binary location or version could not be determined (meaning there was no search for any file).


kapoorlakshya:
Thank you for your contribution!

PR #234, opened 2022-08-03
Download correct geckdriver file on Apple Silicon

stephannv:
Geckodriver splits its Mac OS files into 2 versions (`macos.tar.gz` and `macos-aarch64.tar.gz`), and `webdrivers` doesn't work properly if `macos.tar.gz` is used on Apple Sillicon platform.
Geckodriver releases: https://github.com/mozilla/geckodriver/releases

### DIFF
--- a/lib/webdrivers/geckodriver.rb
+++ b/lib/webdrivers/geckodriver.rb
@@ -51,7 +51,7 @@ module Webdrivers
         when 'linux'
           "linux#{System.bitsize}.tar.gz"
         when 'mac'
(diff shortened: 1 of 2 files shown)


kapoorlakshya:
@stephannv Thanks for your contribution! Can you please add a test for this and rebase on `main`? I just fixed two broken tests and CI checks should be green now.

stephannv:
Done @kapoorlakshya.

stephannv:
Specs fixed.

Issue #237, opened 2022-09-28
Webdrivers::NetworkError, probably due to new chromedriver naming


ansonhoyt:
#### Summary
It looks like Webdrivers' `download_file` gets a 404 because Google changed chromedriver's filename scheme with 106.0.5249.61.

#### Debug Info
* Webdrivers version: 5.1.0
* Ruby version: 3.0.4 (and 2.7.6 with another app. Both compiled for Apple's arm64, *not* x86_64 + Rosetta.)
* Operating system / CI Environment: macOS 12.6 with an M1 Max
* Browser and version: chromedriver 106.0.5249.61

#### Expected Behavior
`download_file` should download: `https://chromedriver.storage.googleapis.com/106.0.5249.61/chromedriver_mac_arm64.zip`

#### Actual Behavior

`download_file` tried to download: `https://chromedriver.storage.googleapis.com/106.0.5249.61/chromedriver_mac64_m1.zip`
```
/Users/ahoyt/.rvm/gems/ruby-3.0.4@eita-activity/gems/webdrivers-5.1.0/lib/webdrivers/network.rb:19:in `get': Net::HTTPServerException: 404 "Not Found" with https://chromedriver.storage.googleapis.com/106.0.5249.61/chromedriver_mac64_m1.zip (Webdrivers::NetworkError)
	from /Users/ahoyt/.rvm/gems/ruby-3.0.4@eita-activity/gems/webdrivers-5.1.0/lib/webdrivers/system.rb:74:in `block (2 levels) in download_file'
	from /Users/ahoyt/.rvm/rubies/ruby-3.0.4/lib/ruby/3.0.0/tempfile.rb:317:in `open'
	from /Users/ahoyt/.rvm/gems/ruby-3.0.4@eita-activity/gems/webdrivers-5.1.0/lib/webdrivers/system.rb:73:in `block in download_file'
	from /Users/ahoyt/.rvm/gems/ruby-3.0.4@eita-activity/gems/webdrivers-5.1.0/lib/webdrivers/system.rb:72:in `chdir'
	from /Users/ahoyt/.rvm/gems/ruby-3.0.4@eita-activity/gems/webdrivers-5.1.0/lib/webdrivers/system.rb:72:in `download_file'
	from /Users/ahoyt/.rvm/gems/ruby-3.0.4@eita-activity/gems/webdrivers-5.1.0/lib/webdrivers/system.rb:63:in `download'
[...]

PR #240, opened 2022-09-30
Add tests for chromedriver on apple machines

jmccure:
Adds unit tests for https://github.com/titusfortner/webdrivers/pull/239

PR #241, opened 2022-10-08
Fixes downloading on m1 macs from version 106.0.5249.61 onwards

frenesim:
Fix a similar Issue to #237

The download is failing again for the newer chrome version. From version 106.0.5249.61 onwards seems the filename is 'chromedriver_mac_arm64.zip'.

https://chromedriver.storage.googleapis.com/

Let me know if we should open an issue to this one.

I found testing this change a bit hard, so I went on testing a private method 💀 . I tested the method I perceive as the best candidate to be the public API of a new class.

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -97,11 +97,11 @@ module Webdrivers
         false
       end
 
-      def apple_filename(driver_version)
+      def apple_m1_filename(driver_version)
         if apple_m1_compatible?(driver_version)
-          driver_version >= normalize_version('106.0.5249.61') ? 'mac_arm64' : 'mac64_m1'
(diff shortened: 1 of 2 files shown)


jmccure:
@frenesim I don't think this really changes anything. The only change is the method name and changing the strings to double quotes. What error are you fixing? The two latest chromedriver versions have the same filename suffix.

`106.0.5249.61/chromedriver_mac_arm64.zip`
`107.0.5304.18/chromedriver_mac_arm64.zip`

Are you using the latest version of this gem?

Also I have tests that don't test the private methods in https://github.com/titusfortner/webdrivers/pull/240

frenesim:
You are right. I thought I had the latest version, but no.
Funny, we end up with almost the same code. :)

PR #249, opened 2023-07-27
Fixed Webdrivers::VersionError with chrome version greater than 115

sadahiro-ono:
similar Issue: https://github.com/titusfortner/webdrivers/issues/247

f the version is 115 or higher, the API can be used, so if the version is 115 or higher, the stable version is obtained and downloaded from the API.

refarence
- https://chromedriver.chromium.org/downloads
- https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints

### DIFF
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -3,6 +3,7 @@
 require 'shellwords'
 require 'webdrivers/common'
 require 'webdrivers/chrome_finder'
+require 'json'
 
 module Webdrivers
   class Chromedriver < Common
@@ -63,7 +64,7 @@ module Webdrivers
       private
 
       def latest_point_release(version)
-        normalize_version(Network.get(URI.join(base_url, "LATEST_RELEASE_#{version}")))
+        normalize_version(stable_version) || normalize_version(Network.get(URI.join(base_url, "LATEST_RELEASE_#{version}")))
       rescue NetworkError
         msg = "Unable to find latest point release version for #{version}."
         msg = begin
@@ -106,7 +107,11 @@ module Webdrivers
       end
 
       def direct_url(driver_version)
-        "#{base_url}/#{driver_version}/chromedriver_#{driver_filename(driver_version)}.zip"
+        if normalize_version('115') > driver_version
+          direct_url_for_over_115(driver_version)
+        else
+          "#{base_url}/#{driver_version}/chromedriver_#{driver_filename(driver_version)}.zip"
+        end
       end
 
       def driver_filename(driver_version)
@@ -148,6 +153,22 @@ module Webdrivers
       def sufficient_binary?
         super && current_version && (current_version < normalize_version('70.0.3538') ||
           current_build_version == browser_build_version)
+      end
+
+      def chrome_for_testing_base_url
+        'https://googlechromelabs-github-io.translate.goog/chrome-for-testing'
+      end
+
+      def stable_version
+        uri = URI.join(chrome_for_testing_base_url, 'last-known-good-versions.json')
+        res = Network.get(uri)
+        JSON.parse(res, symbolize_names: true).dig(:channels, :Stable, :version)
+      end
+
+      def direct_url_for_over_115(driver_version)
+        uri = URI.join(chrome_for_testing_base_url, 'last-known-good-versions-with-downloads.json')
+        json = JSON.parse(Network.get(uri), symbolize_names: true).dig(:channels, :Stable, :downloads, :chrome)
+        json.find { |e| e[:platform] == driver_filename(driver_version) }[:url]
       end
     end
   end
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -159,7 +159,7 @@ module Webdrivers
       end
 
       def stable_version
-        uri = URI.join(chrome_for_testing_base_url, '/chrome-for-testinglast-known-good-versions.json')
+        uri = URI.join(chrome_for_testing_base_url, '/chrome-for-testin/last-known-good-versions.json')
         res = Network.get(uri)
         JSON.parse(res, symbolize_names: true).dig(:channels, :Stable, :version)
       end
--- a/lib/webdrivers/chromedriver.rb
+++ b/lib/webdrivers/chromedriver.rb
@@ -63,7 +63,7 @@ module Webdrivers
       private
 
       def latest_point_release(version)
-        normalize_version(stable_version) || normalize_version(Network.get(URI.join(base_url, "LATEST_RELEASE_#{version}")))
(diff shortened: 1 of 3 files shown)


titusfortner:
Thanks for the PR. It looks like it isn't backwards compatible. Can you look at why the tests are failing?

sadahiro-ono:
@titusfortner 
Local tests passed, but I made some fixes to lower the impact, could you please run a CI?

sadahiro-ono:
@titusfortner 
The main branch is also not passing the tests. I sorry, Could you please confirm this for me?
https://github.com/titusfortner/webdrivers/actions/runs/5683791993/job/15405104246

alexshenia:
` expected StandardError with message matching /Net::HTTPServerException: 404 "Not Found"/, got #<Webdrivers::NetworkError: Net::HTTPClientException: 404 "Not Found" with https://github.com/mozilla/geckodriver/releases/download/v0.2.0/geckodriver-v0.2.0-linux64.tar.gz>`

Just need to change from `StandardError` to `Webdrivers::NetworkError`

alexshenia:
` Webdrivers::NetworkError:
       Net::HTTPClientException: 404 "Not Found" with https://msedgedriver.azureedge.net/114.0.1823.90/edgedriver_mac64.zip`
https://msedgedriver.azureedge.net/ does not have this build  `114.0.1823.90/edgedriver_mac64.zip`

titusfortner:
StandardError is the superclass, it is the value of the String it failed on. Not sure if it is happenstance or common that MS doesn't do releases for all platforms for all patch versions, but not good either way.

lib/webdrivers/chromedriver.rb
```diff
@@ -77,8 +81,14 @@ def latest_point_release(version)
           "#{msg} A network issue is preventing determination of latest chromedriver release."
         end
 
+        url = if version >= normalize_version('115')
```
titusfortner, line 84: We can use https://chromedriver.chromium.org/downloads/version-selection for the error message

lib/webdrivers/chromedriver.rb
```diff
@@ -105,8 +115,17 @@ def apple_filename(driver_version)
         end
       end
 
+      def apple_filename_for_api(driver_version)
+        if apple_m1_compatible?(driver_version)
+          driver_version >= normalize_version('106.0.5249.61') ? 'mac-arm64' : 'mac64-m1'
```
titusfortner, line 120: if we put this conditional in `#apple_filename` we don't need to add `#apple_filename_api`.

lib/webdrivers/chromedriver.rb
```diff
@@ -149,6 +172,33 @@ def sufficient_binary?
         super && current_version && (current_version < normalize_version('70.0.3538') ||
           current_build_version == browser_build_version)
       end
+
+      def chrome_for_testing_base_url
```
titusfortner, line 176: Not sure if it matters, but this breaks https://github.com/titusfortner/webdrivers/wiki/Using-with-VCR-or-WebMock

titusfortner (changes requested):
1. Thanks for putting this together and adding tests
2. I really hate looking at old code I've written. 
3. We probably need to exclude `chromedriver.rb` from `Metrics/ClassLength` in `.rubocop.yml`.

lib/webdrivers/chromedriver.rb
titusfortner: hmm, I deleted the wrong comment looks like. Let me fix.

lib/webdrivers/chromedriver.rb
```diff
@@ -115,7 +134,11 @@ def driver_filename(driver_version)
         elsif System.platform == 'linux'
           'linux64'
         elsif System.platform == 'mac'
-          apple_filename(driver_version)
+          if driver_version >= normalize_version('115')
```
titusfortner, line 137: If we move this conditional to `#apple_filename` we don't need to add `#apple_filename_api`.

lib/webdrivers/chromedriver.rb
titusfortner: moved the comment to where it was supposed to go below.

spec/webdrivers/chromedriver_spec.rb
```diff
@@ -66,7 +66,7 @@
         allow(Net::HTTP).to receive(:get_response).and_raise(SocketError)
         allow(chromedriver).to receive(:exists?).and_return(false)
 
-        msg = %r{Can not reach https://chromedriver.storage.googleapis.com}
+        msg = %r{Can not reach https://googlechromelabs.github.io/chrome-for-testing/latest-patch-versions-per-build.json}
```
yykamei, line 69: How about considering the both cases for v115+ and v114 or earlier? The Gem is expected to print "Can not reach ...", and the target URL is not important here.

```suggestion
        msg = %r{Can not reach (?:https://googlechromelabs.github.io|https://chromedriver.storage.googleapis.com)}
```

spec/webdrivers/chromedriver_spec.rb
```diff
@@ -101,7 +101,7 @@
       it 'raises ConnectionError if offline' do
         allow(Net::HTTP).to receive(:get_response).and_raise(SocketError)
 
-        msg = %r{Can not reach https://chromedriver.storage.googleapis.com/}
+        msg = %r{Can not reach https://googlechromelabs.github.io/chrome-for-testing/latest-patch-versions-per-build.json}
```
yykamei, line 104:
```suggestion
        msg = %r{Can not reach (?:https://googlechromelabs.github.io|https://chromedriver.storage.googleapis.com)}
```

spec/webdrivers/chromedriver_spec.rb
```diff
@@ -223,11 +241,12 @@
     it 'raises ConnectionError when offline' do
       allow(Net::HTTP).to receive(:get_response).and_raise(SocketError)
 
-      msg = %r{^Can not reach https://chromedriver.storage.googleapis.com}
+      msg = %r{^Can not reach https://googlechromelabs.github.io/chrome-for-testing/latest-patch-versions-per-build.json}
```
yykamei, line 244:
```suggestion
      msg = %r{Can not reach (?:https://googlechromelabs.github.io|https://chromedriver.storage.googleapis.com)}
```

lib/webdrivers/chromedriver.rb
sadahiro-ono: Do we need to support a version that we cannot get from[ this endpoint](https://googlechromelabs.github.io/chrome-for-testing/latest-patch-versions-per-build.json)?
I believe we can change the base_url to the API endpoint if we don't need it.

titusfortner: yes, we still need to support the older versions

titusfortner:
Merged here — https://github.com/titusfortner/webdrivers/commit/97fb1e2b842c427a885553ac695ecdbf303344ba

Thank you!

PR #256, opened 2023-08-15
update selenium-webdriver to account for latest version

zmariscal:
The gem `selenium-webdriver` has been updated:
https://github.com/SeleniumHQ/selenium/blob/trunk/rb/CHANGES#L1

To ensure we can update this gem in or project we need to update `webdrivers.gemspec` to ensure we can get the latest version.

Resolves issue: https://github.com/titusfortner/webdrivers/issues/257


titusfortner:
Thanks for the PR, but see the comment on the related issue.

